PR #25, opened 2020-07-22
hd video

tommhendrick:
This allows hd video and audio that can play on omxplayer by omitting webm format

### DIFF
--- a/karaoke.py
+++ b/karaoke.py
@@ -288,7 +288,7 @@ class Karaoke:
         logging.info("Downloading video: " + video_url)
         dl_path = self.download_path + "%(title)s---%(id)s.%(ext)s"
         cmd = [self.youtubedl_path,
-        	'-f', 'mp4',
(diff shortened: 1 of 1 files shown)


tommhendrick:
I believe it combines them into one file after but not sure if it uses ffmpeg. I didn't have to install anything extra on the PI for it to work.

vicwomg:
When I tried this, I got two separate files. Is there a missing parameter, like --merge-output-format mp4  ?

tommhendrick:
just checked. it uses FFMPEG to combine them. however on my rpi3 b+ I never manually installed it so maybe it came with something. So I guess the solution to hd video and audio is to include ffmpeg in the install script if you truly want HD. and then use my line above.

-best will only find the best single file which usually isn't HD.

vicwomg:
manually merged this in. setup.sh now includes ffmpeg and passing --high-quality to app.py will utilize the following quality specification to the -f option in youtube-dl: 

```bestvideo[ext!=webm][height<=1080]+bestaudio[ext!=webm]/best[ext!=webm]```

PR #28, opened 2020-07-30
UI Customization

petemarvin:
- Improve Remove "karaoke" references function (i.e. KARAOKE VERSION, HD, capitalize each word) still planning to improve more on this.
- Added Fontello customized icons and replace the Unicode Hex Character Code to Fontello Icons. (problem with different icons on Android/IOS/Desktop)
- Added soundbar GIF to indicate that the song is playing in the Now Playing Header
- Added **is_pause** function. Hides the Player Control Box when there is no song playing. Show pause ❚❚ when the song is playing and show play ▶️  when the song is paused.

BTW, nice work on the app. Been looking for an app like this.


vicwomg:
Nice work! I'd prefer to keep the logo unchanged though, so if you can remove that I'll consider the merge.

templates/edit.html
```diff
@@ -6,9 +6,26 @@
   
     $('#remove-karaoke').click(function(e) {
         var new_name = $('#new_file_name').val();
-        new_name = new_name.replace(/\([Kk][Aa][Rr][Aa][Oo][Kk][Ee].*\)/g, "");
-        new_name = new_name.replace(/[Kk][Aa][Rr][Aa][Oo][Kk][Ee]/g, "");
```
vicwomg, line 10: What is the purpose of the whitespace replacements in line 10 and 15?

templates/base.html
```diff
@@ -39,18 +42,18 @@
...
+              Now Playing
             </a>
 
             <a class="navbar-item" href="{{ url_for('queue') }}">
               Queue
             </a>
             <a class="navbar-item" href="{{ url_for('search') }}">
-              Songs
+              Search Songs
             </a>
```
vicwomg, line 57: The titles were deliberately small to fit on iPhone 5 resolutions, otherwise they start to line break. 

Also, please do not change the logo to a personal logo in a master pull request (line 45)

templates/edit.html
petemarvin: There were times after removing the words KARAOKE... there were some white spaces left. Just a quick fix.

templates/base.html
petemarvin: apologies I'll just have to revert is back.

karaoke.py
```diff
@@ -483,6 +487,10 @@ def pause(self):
         if (self.is_file_playing()):
             logging.info("Pausing: " + self.now_playing)
             self.process.stdin.write("p")
```
vicwomg, line 489: Nice.

```self.is_pause = not self.is_pause``` could be a nice one-liner for 490-493

File: karaoke.py [hunk 4]
```diff
@@ -198,16 +200,16 @@ def render_splash_screen(self):
             if (not self.hide_ip):
                 p_image = pygame.image.load(self.generate_qr_code())
                 p_image = pygame.transform.scale(p_image, (150, 150))
-                self.screen.blit(p_image, (0,0))
```
vicwomg, line 201: are you finding that there are screen cropping issues? Is it possible you have a miscalibrated screen?

templates/edit.html
```diff
@@ -6,9 +6,26 @@
...
+        new_name = new_name.replace(/[Kk][Aa][Rr][Aa][Oo][Kk][Ee]/g, " ");
+        new_name = new_name.replace(/\s\s+/g, " ");
+        new_name = new_name.replace("- -", "-");
+        new_name = new_name.replace("--", "-");
+        new_name = new_name.replace("as popularized by", "");
         new_name = new_name.trim();
+        
+        
+        
+        //title case
```
vicwomg, line 23: hmm, I feel like if we're just going to titlecase the end might as well toLowerCase() before all the replace calls, to simplify them (wouldn't need to do this [Kk][Aa][Rr][Aa][Oo][Kk][Ee] business and could do simply: "karaoke" to be more readable.

karaoke.py
petemarvin: Maybe it was just my screen.

templates/edit.html
petemarvin: I agree. Unfortunately, I did commit to my own preferences.

vicwomg:
Overall, the non-emoji icons and pause/play functionality are awesome. Thanks for doing that. I'd love to get this into master. Main thing holding is the personal logos.

Things are looking a little wacky on the queue view with the ordering icons, since they are not fixed width, but I may be changing this to a table in a future rev to make it look cleaner anyway

petemarvin:
Apologies I'm new to this Github, I have pulled a request that was based on my preferences. Anyways, I really love your app maybe you can consider this as a suggestion only. I trust that you are more knowledgeable on this matter.

petemarvin:
Please refer to #29

PR #30, opened 2020-08-01
Last.fm API to make suggestion for "title - artist" filename

petemarvin:
Added, "Suggest from Last.fm" (last.fm API) to make the search for "title - artist" suggestions for filename and improve the karaoke reference removal. Demo https://jsfiddle.net/petemarvin/n7p8c34b/169/

### DIFF
--- a/templates/edit.html
+++ b/templates/edit.html
@@ -2,31 +2,19 @@
 
 {% block scripts %}
 <script>
-  $( function() {
+ $( function() {
   
     $('#remove-karaoke').click(function(e) {
-        var new_name = $('#new_file_name').val();
-        new_name = new_name.replace(/\([Hh][Dd][ ].*\)/g, " ");
-        new_name = new_name.replace(/\s\s+/g, "");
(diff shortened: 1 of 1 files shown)


vicwomg:
Very cool!

I actually recently did some clean up worn on the "remove karaoke" function and renamed it to "auto-format". So I may just cut and paste the last.fm function and add it in manually. Is that your personal API key? I'm not sure we should be checking in API keys to the codeline, otherwise people could steal them. Would have to think about the best way to do this. Perhaps you'd need to supply your own with a config file.

petemarvin:
Yeah, It is my personal API key. Maybe perhaps you could add in the info.html an input field to be filled with client's own API? or maybe a dedicated PiKaraoke last.fm account. 

Anyways thanks for the updates.

vicwomg:
Merged into develop. I made a modification to call clean_title on the string before passing it to last.fm, since I found that having junky text in there negatively affected the returned search results. Great work!

PR #34, opened 2020-08-04
Added YouTube images thumbnail based on selected search result

petemarvin:
The thumbnail will help the "singer" find his/her preferred video. Prevents from choosing a video that is only static or blank screen.


vicwomg:
looks great! Merging to development for a closer look

PR #35, opened 2020-08-09
Prevent overwrite if filename exist

petemarvin:
This will prevent overwriting an existing song filename.


vicwomg:
looks great, thank you!

PR #37, opened 2020-08-19
Song Search Page

petemarvin:
- By default guest always go directly to _Add New Song_ section to search songs without checking first if the song is already downloaded. Chances are songs will have multiple copies. To prevent this, the guest will only use one input field to type the song search keyword. If search keyword is available the select (dropdown) will show the available songs. The guest can click the available song and the queue button will show. If search keyword is unavailable the guest can click the search button to trigger the search function.
- Since I was not familiar with jquery-UI, I've change to another plugin "selectize". Problem I had with jquery-UI combobox was the search keyword function. In jquery-UI, it will search the whole inputted text wheareas selectize will search the inputted text anywhere in the song filename.
Screenshot_2020-08-19-13-11-12-02 Screenshot_2020-08-19-13-11-56-82

- Added some UI enhancements.


vicwomg:
Looks awesome.

vicwomg:
I think you forgot to check-in the selectize css and js files, but I added them. Also did some minor design/text tweaks and fixed the issue where clicking the up arrow doesn't close the dropdown. on develop branch now. Thanks for this! I definitely think it improves things around search.

PR #44, opened 2020-08-29
Options to use VLC Player

petemarvin:
To use VLC player add argument `sudo python app.py --use-vlc`
Tested in RPI 3 B+ and Raspbian Desktop (Virtual Box).
VLC player is useful without or cannot install OMXplayer.


vicwomg:
wow. This is awesome. This opens the doors to running on OSX and Linux. I'll check it out

PR #57, opened 2020-10-31
Develop fixyt

### LINKED ISSUE #54: Search via youtube-dl appears to be broken
Search of content not previously stored has stopped working and updating youtube-dl isn't fixing it.
A bit of digging reveals that the hosting repository at yt-dl.org is currently pointing to a GitHub DMCA take-down notice.

EdFabre:
Updated setup-pi.sh for raspberry-pi's and README.md for other users to explain how to install correct version of youtube-dl. During my testing I noticed youtube-dl Download was also broken. The new version is tested and working.

Added Features
* Search by URL
* Search by Raw String by including "raw" in search
* Search for lyric video's by including "lyrics" in search
* Set [youtube-scrape](https://github.com/HermanFassett/youtube-scrape) as an internal NodeJS package that automatically executes if youtube-dl throws exceptions. 
Fixes #54

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ __pycache__/
 songs/
 qrcode.png
 .DS_Store
+youtube-scrape/
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Karaoke song search and queueing system that uses Youtube videos as a source. ~F
 
 If you want to support this project with a little monetary tip, it's much appreciated: https://www.paypal.me/paniquejazz
 
-## What's new (September 2020)
+## What's new (October 2020)
 
 - VLC support! Omxplayer is still default on raspberry pi, which can use both.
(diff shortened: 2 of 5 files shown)


vicwomg:
Hey EdFabre, really appreciate your contribution! But now that the dust settled on searching, I think we should separate these things out into separate issues and pull requests.

I put search by URL in: #59 

Regarding search "raw" and "lyrics" searching, the problem with detecting strings in the search field is the song title might legitimately contain those strings. "Raw" for example. 

It seems to me like the more flexible solution is just to allow the user to include non-karaoke matches. Perhaps a checkbox on the UI that would remove the forced "karaoke" string in the query. That way folks can use the search to download whatever they want, really.  Filed as #60

EdFabre:
@vicwomg I agree, I will close out my request. 

In other news, I actually took the liberty of creating a setup-win.exe script to setup pikaroke for Windows users, similar to how setup-pi.sh does. I will put it in my fork for your review.

PR #63, opened 2020-11-04
Readme patch 1

Hartebee5t:
First attempt at a mainline pull request.
Corrects some mismatched entries between options-list and option descriptions, plus a couple of typos in paragraphs following the options list.
There’s also a suggested addition to the description of the Home screen, but I’m not sure I’ve worded it consistently with the rest of the descriptions in that section.


Hartebee5t:
Hmmm, noob anomaly...
 I’ve just noticed this PR contains 5 commits but the first 4 were to bring my master into sync with your last Readme update of the previous day ([1ae58da](https://github.com/vicwomg/pikaraoke/commit/1ae58da9fe76622808a884f7094f3ff6a88d67cc)), so those four are ineffective here - I think. I had assumed the comparison against your latest commit would produce a PR from that point.
How should I have generated this to avoid the unnecessary first four commits?

vicwomg:
Thank you for your contribution! And sorry for the delay. I have been.... distracted :)

I'm actually not sure the exact nature of those merge commits, but I think they are normal. My guess is your working branch was behind master. If you want to avoid that sort of thing make sure you have the latest pull from master before you branch off and commit any new changes.

PR #99, opened 2021-02-03
Add option to use WAN IP instead of LAN

jramboz:
For those who don't want to give out their wifi password, I've added an option to get and display your external (WAN) IP address instead of the local (LAN) one. Currently this will require users to manually make sure the port is open and directing to the machine running pikaraoke. Next I'll work on automatic port mapping using UPnP.

Pikaraoke first attempts to get the external IP address using UPnP. If that fails, it uses a web call to [ident.me](http://ident.me). If it still can't get a valid external IP, it defaults back to the LAN IP.

This adds an additional requirement -- module `miniupnpc` -- which I've also added to `requirements.txt`.

I'm only a mediocre coder, and this is my first open source pull request. I'm open to suggestions on better ways to do this! :)

### DIFF
--- a/app.py
+++ b/app.py
@@ -612,6 +612,12 @@ if __name__ == "__main__":
         required=False,
     )
     parser.add_argument(
+        "--use-external-ip",
+        action="store_true",
+        help="Use external (wan/internet) ip address instead of local (lan) ip",
+        required=False,
+    )
+    parser.add_argument(
         "--hide-ip",
         action="store_true",
         help="Hide IP address from the screen.",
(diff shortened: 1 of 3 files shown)


vicwomg:
Oh, sorry I never responded to this. I hadn't had time to fully review. My main reservations were whether it works when an ipv6 is returned. 

I can certainly see the use case for pikaraoke in public places, but I'd also be concerned about overall security. But I guess that could be left up to the administrator

jramboz:
Yeah, I canceled because I wanted to rethink it a bit and test with IPv6. Going to work on it a bit further and resubmit to the development branch.

Issue #108, opened 2021-02-15
Recursively search subfolders within download directory


jramboz:
For those of us who obsess over organizing our files and/or already have large MP3+G libraries sorted into subfolders.

vicwomg:
Actually, merged that but it's not working for me.

- created subdir  in my song folder ./sub
- added files 000000.mp4 and zzzzzzz.zip

Neither shows up in browse. I'm testing on OSX, would also be good to know if Windows works with this too. Also want to confirm that files are sorted properly (filenames in subfolders should be sorted with the general list)

jramboz:
Very strange. It worked for me on both macOS and Raspbian. Let me go back and double check to see if I'm missing something. What version of python are you using? I think the recursive feature of glob were added in 3.5.

jramboz:
I just tested again and confirmed it's working for me on macOS and Raspbian. Songs are sorted correctly by filename, regardless of which folder they're in. I'm updating my Windows VM and then I'll test there as well. Other than the python version, I have no idea why it wouldn't be working on your side, but I'll keep digging.

vicwomg:
I'm on 3.9

```
vicwong@vics-mbp:~/pikaraoke-songs $ python3 --version
Python 3.9.1
```

vicwomg:
oh, I'm sorry. I am seeing them. But as I feared they are not being sorted correctly. Since the song list is sorted by path, songs in ./sub will be sorted with the "s" section. You may need to look into the method for get_available_songs and make sure it sorts by the filename... which could be tricky

jramboz:
Ah ha. Alright, let me look into that. If you don't mind, I'll close this issue and open a new one for the bug?

vicwomg:
sure thing

PR #109, opened 2021-02-15
#108 Recursively search subfolders for songs

jramboz:
Adds enhancement #108


vicwomg:
One-liners are always nice :)

vicwomg:
actually, doesn't seem to work for me on closer look. See comment in #108

Issue #110, opened 2021-02-17
Bug: Files in subfolders not sorting correctly


jramboz:
Files in subfolders are being sorted by the subfolder name rather than by filename. So for example, any files in `songs/sub` will appear sorted under S.

jramboz:
I think the trick here is going to be to create a custom function for they `key` parameter in the `sorted()` call. We need to look for a `/` character in the string and, if it's there, lop off everything up to and including the last `/`. Then return `str.lower()` on what's left.

I'll put this together and test it tonight or tomorrow.

jramboz:
Haha. `os.path.basename()`, of course.

PR #111, opened 2021-02-18
#110 Fix sorting for files from subfolders

jramboz:
Another one line fix!


vicwomg:
move over Rodney Dangerfield, we've got a new king of one-liners

PR #159, opened 2021-12-14
Localize PiKaraoke and include Chinese translations

tbelaire:
I marked up the entire app with {% trans %} markers and _() calls, and 

I made sure the app can install cleanly in a virtualenv

```shell
$ python -m venv ./.venv
$ source .venv/bin/activate
$ pip install -i requirements.txt
```

My editing loop is something like this:
```shell
$ vim templates/*.html
$ pybabel extract -F babel.cfg -o messages.pot --add-comments="MSG:" --strip-comment-tags  --sort-by-file . && pybabel update -i messages.pot -d translations
$ vim translations/*/messages.po
$ pybabel compile -d translations/ && python app.py -y /usr/bin/youtube-dl --hide-splash-screen
```

I had to change some wordings in the UI, as for example, "Queue" (noun, title of page) and "Queue" (verb, as in add to queue) are both spelled exactly the same, but need different translations.  The limitations of gettext's approach is that translations are uniquely keyed by just the English string, so it's impossible to give that two translations.  I added up changing it to "Add to queue" instead.


akhepcat:
queue being synonymous with enqueue is silly sometimes.   inflammable / flammable  is worse, though.  lol

tbelaire:
Ok I got to test on a pi, and the setup pi script doesn't have Flask-Babel.
Also the new pygame version doesn't work right away, I need to do this trick

https://forums.raspberrypi.com/viewtopic.php?t=270226

But I'm having trouble getting it to boot at launch with that, so I need to work on that more.

vicwomg:
If I recall correctly the latest pygame on requires some sdl libraries to
be installed by apt, they are in the requirements.txt in one of the
branches here for pi4 support

vicwomg:
You might want to try the following to get the latest pygame working, which I added to the 160 branch setup-pi.sh script:

`sudo apt-get install libjpeg-dev libsdl2-image-2.0-0 vlc python3-pip ffmpeg libsdl2-ttf-dev -y`

tbelaire:
Alright, I can confirm the translations are working on a Pi 0W.  

I'll clean up this to only have the very bare minimum for translations, but I have a few other changes I will open a second pull request for.

I got it working well on boot as a systemd service, and by doing so I can also avoid running as root and needing to patch vlc, which I like.

tbelaire:
There, this is super clean now, and ready to merge in.  

If you want, I can also not commit the .mo files, and have those get compiled as needed, since committing binary files isn't exactly great, but I wanted it to work if you just pulled and ran it, no build step.

vicwomg:
excellent. Sorry for the delay in getting around to testing all this. Looks good to me. I will merge and make an announcement on the next release

PR #166, opened 2021-12-24
Requesting pull for download speed fix.

cockernutx:
Changed from youtube_dl library to yt_dlp for a faster download speed.


friki67:
Hello! this breaks #167

vicwomg:
Thank you! This was addressed in a separate PR and merged to master

PR #174, opened 2022-01-02
Fix bullseye / pi4 setup and config

vicwomg:


Romeo1984:
Additional changes need to be made:
***Comment out pygame Darwin install in requirements.txt
--pygame; sys_platform != 'darwin'
--pygame==2.1.0; sys_platform == 'darwin'
++pygame==2.1.0

Add apt-get pygame uninstall in setup-pi.sh

echo "Uninstalling old pygame versions..."
++sudo apt-get remove python3-pygame -y

PR #185, opened 2022-01-21
2 useful major improvements

xuancong84:
Improvements:
1. Stream both video and audio to HTTP, so any web browser on the local area network can also open the Karaoke TV display
2. Songs with non-English characters in their filenames are now searched and sorted according to English transliteration, with leading non-alphanumeric characters removed, e.g., "あの日から.mp4" (in Japanese, transliteration: "a no ri ka ra") is grouped under A; "《歌声与微笑》.mp4" (in Chinese, transliteration: "ge sheng yu wei xiao") is grouped under G.
3. You can now launch the system remotely via SSH
4. You can now drag a song up and down the play queue
5. Added audio delay into player control, because YouTube songs have synchronized lyric, this makes singing difficult as there is not enough time to look at the lyric

Take note:
- Launch the Karaoke system by running ./run.sh, now all services are launched in a `tmux` session, so `tmux` must be installed
- Anaconda3 must be installed in $HOME/anaconda3 folder, with all requirements installed


vicwomg:
Hi, xuancong84 I appreciate your updates and would like to review them for merging to master, but your formatter has set all the indentations in python to tabs, making it very hard to review the diff file. Can you change this to use 4 spaces instead of TABs?

xuancong84:
Thanks @vicwomg . I think I will stick to tabs indentation because:
1. the tab character is only 1 byte while 4 space characters are 4 bytes, so it makes the source code smaller and execution/interpretation faster (though the difference is small)
2. all modern IDEs like PyCharm, Eclipse, Visual Code, etc., they support adjustable indentation for the Tab indentation, i.e., you can make the tab character appear as 1/2/3/4/etc spaces in IDE without changing the source code
3. VIM also support tab indentation, you may refer to my repo for the .vimrc (https://github.com/xuancong84/linux-home.git ) or google yourself

You may want to change indentation for all your source code except those web scripts, then perform a merge. Or you can just keep this PR open and let me merge your future updates. Many thanks! -:)

PR #224, opened 2022-12-19
Suppress FileNotFoundError when deleting songs

alcalda:


vicwomg:
Thanks for your contribution!

PR #285, opened 2023-12-21
workaround for 'required interaction' with browser

RecursiveFunctions:
this allows PiKaraoke to autoplay without needing to click 'Confirm'. This only applies to the browser on the host machine. visiting /splash from your phone would still require tapping 'Confirm'.

This works with raspberry pi4 (8GB ram) with bookworm

### DIFF
--- a/app.py
+++ b/app.py
@@ -867,6 +867,7 @@ if __name__ == "__main__":
                 service = None
             options = Options()
             options.add_argument("--kiosk")
+            options.add_argument("--start-maximized")
             options.add_experimental_option("excludeSwitches", ['enable-automation'])
             driver = webdriver.Chrome(service=service, options=options)
             driver.get(f"{k.url}/splash" )


vicwomg:
Are you sure this is necessary? Selenium auto-dismisses the Confirm button, so this should be already handled. The only time you need to click it is you launched manually in another browser.

RecursiveFunctions:
Ah, didn't know that. But this line also fixes an unrelated issue where, in a wayland environment the browser window is displayed improperly, so there is that.

I happened to be in wayland, just tested it both with and without the change to be sure.

vicwomg:
Oh good to know I’ll merge this then! I saw that Wayland issue and I just
switched out of Wayland in raspi config and forgot about it.

vicwomg:
I have manually added this to my working branch, so no need to merge this. Thank you!

vicwomg:
> > this allows PiKaraoke to autoplay without needing to click 'Confirm'. This only applies to the browser on the host machine. visiting /splash from your phone would still require tapping 'Confirm'.
> > This works with raspberry pi4 (8GB ram) with bookworm
> 
> I know the issue is closed but I have to ask, how does --start-maximized bypass the need to click the button?

It doesn't. This fixes another issue " this line also fixes an unrelated issue where, in a wayland environment the browser window is displayed improperly, so there is that"

As for how I bypass it otherwise, Pikaraoke auto-dismisses the confirm button with a simulated button click in selenium. But only if you're running in headed mode. Otherwise, you must manually click it.

Issue #287, opened 2023-12-21
[BUG] -d and -y options don't work with paths that have spaces in them


RecursiveFunctions:
this screencap shows the issue with only -d , but it's the same as with -y
![issue_screencap](https://github.com/vicwomg/pikaraoke/assets/3803876/01304947-3440-4465-8107-97822171d5e6)

I tried with quotes and escaping the space but the path handling for both options does not work with paths that have spaces in the name.

 - OS: Raspberry Pi OS
 - OS version: Raspberry pi Debian Bookworm 64-bit
 - Hardware: Raspberry Pi 4 (8GB)

vicwomg:
Not to self/whoever wants to take this on: Looks like there's some additional handling for python argparse required to parse spaces:

https://stackoverflow.com/a/26990349/2909171

vicwomg:
This should be fixed in  the latest pull of this branch
`277-html_ffmpeg_video_rendering`

lmk if it works out

vicwomg:
merged this to main as 1.3

PR #289, opened 2023-12-23
add songs from external drives to ~/pikaraoke-songs

RecursiveFunctions:
Added 2 scripts that create symbolic links from all songs from external drives to ~/pikaraoke-songs, including songs in subfolders.

### DIFF
--- a/app.py
+++ b/app.py
@@ -57,6 +57,11 @@ def filename_from_path(file_path, remove_youtube_id=True):
             rc = rc.split("---".encode("utf-8", "ignore"))[0]
     return rc
 
+def arg_path_parse(path):
+    if (type(path) == list):
+        return " ".join(path)
+    else:
+        return path
 
 def url_escape(filename):
     return quote(filename.encode("utf8"))
@@ -681,6 +686,7 @@ if __name__ == "__main__":
     parser.add_argument(
         "-d",
         "--download-path",
+        nargs='+',
         help="Desired path for downloaded songs. (default: %s)" % default_dl_dir,
         default=default_dl_dir,
         required=False,
@@ -688,6 +694,7 @@ if __name__ == "__main__":
     parser.add_argument(
         "-y",
         "--youtubedl-path",
+        nargs='+',
         help="Path of youtube-dl. (default: %s)" % default_youtubedl_path,
         default=default_youtubedl_path,
         required=False,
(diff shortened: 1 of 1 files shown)


RecursiveFunctions:
Not sure how to properly make a merge request here but the main changes I wanted to add were the 2 scripts and a dependency (or sudo apt install) for plocate in setup.sh

combine.sh binds mount points to /media/unified using names like sda1 sda2 and such.  This solves issue #287 in a roundabout way.

create_database.sh uses plocate to create a local database with the absolute paths to all songs. 

Still thinkin of a way to use this to make a cache so you don't have to load every song on startup.

combine.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+shopt -s globstar
+
+path="/home/$(logname)/pikaraoke-songs"
```
vicwomg, line 4: Assumes the user is downloading to pikaraoke-songs, which may not be the case if they change the download dir.

combine.sh
```diff
@@ -0,0 +1,42 @@
...
+done
+
+#run plocate and get a list of all song paths in the combined folder
+bash /home/$(logname)/pikaraoke/create_database.sh
+
+#add songs to ~/pikaraoke-songs
+while IFS= read -r song || [ -n "$song" ]; do
+  if [[ ! -e "$path/$(basename "$song")" ]]; then 
+    BASE=$(basename "$song")
+    ln -sf "$song" "/home/pi/pikaraoke-songs/$BASE"
```
vicwomg, line 38: Assumes the user is "pi"

vicwomg:
I appreciate the pull request, but I may need to hold off on this because it has a lot of hard-coded paths and it's not cross-platform.

combine.sh
RecursiveFunctions: Maybe I should have it take in whatever the user sets for the -d option, and ~/pikaraoke-songs otherwise.

In that case, I'd insert a call to the script in app.py. will revise.

RecursiveFunctions:
>  it's not cross-platform. 

I could add a check for the OS like you have elsewhere in your code.

I suppose the right answer is to rewrite it in python so it works on windows too, although I'm not sure how I'd go about it. Maybe run this script if linux/macOS, and then have a separate python implementation for windows?

vicwomg:
Yes to make it fully cross platform, this should be done in python. The scripts are really only for installing dependencies and setting up python environments. 

I'll post some suggestions in #247

Issue #290, opened 2023-12-24
[BUG] setup.sh: E: The repository ... does not have a Release file.


RecursiveFunctions:
```
pi:~/pikaraoke (master) $ ./setup.sh
Are you sure you want to setup PiKaraoke? (y/n): y
Client is a Debian-based system. Installing binaries

*** RUNNING APT-GET UPDATE ***
Hit:1 http://deb.debian.org/debian bookworm InRelease
Get:2 http://deb.debian.org/debian-security bookworm-security InRelease [48.0 kB]
Get:3 http://deb.debian.org/debian bookworm-updates InRelease [52.1 kB]        
Ign:4 https://cli.github.com/packages bookworm InRelease                       
Hit:5 http://archive.raspberrypi.com/debian bookworm InRelease                 
Err:6 https://cli.github.com/packages bookworm Release
  404  Not Found [IP: 185.199.108.153 443]
Get:7 http://deb.debian.org/debian-security bookworm-security/main arm64 Packages [127 kB]
Get:8 http://deb.debian.org/debian-security bookworm-security/main armhf Packages [125 kB]
Reading package lists... Done                           
E: The repository 'https://cli.github.com/packages bookworm Release' does not have a Release file.
N: Updating from such a repository can't be done securely, and is therefore disabled by default.
N: See apt-secure(8) manpage for repository creation and user configuration details.
ERROR: 'apt-get update' failed with error code: 0


```
**To Reproduce**
-The user must have installed at least one package from a repository that does not have a release file (in this case cli.github.com/packages bookworm release)

-run setup.sh 

**Platform :**
 - OS: Raspberry Pi OS
 - OS version: Raspberry pi Debian Bookworm 64-bit
 
Error is the same whether you run it as user or root.

RecursiveFunctions:
`sudo apt install --only-upgrade ffmpeg chromium-browser chromium-chromedriver -y
`
ought to fix it

vicwomg:
Thanks for looking into this! Can you explain how it works? I'm  admittedly not well versed in the differences between apt-get and apt. My understanding is apt-get update will fetch all the most recent versions from the repositories. 

If that is removed, how does apt know that it is fetching the latest version?

RecursiveFunctions:
It doesn't, you're right. Actually I think the error handling might've been the true culprit here. ideally when running setup.sh it shouldn't exit out just because some unrelated package doesn't have a release file

vicwomg:
[...]

PR #291, opened 2023-12-24
Bug fix: #290 run apt only on required packages

RecursiveFunctions:
fixes a case #290  where the user has a package that triggers this error:
`E: The repository '...' does not have a Release file
`
when running `sudo apt-get upgrade`, by upgrading / installing only what pikaraoke needs

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -9,6 +9,9 @@ if [ $REPLY = "y" ]
 if [[ $(cat /etc/os-release | grep -i debian) != "" ]]; then
   echo "Client is a Debian-based system. Installing binaries"; 
   echo
+  echo "*** RUNNING APT-GET UPDATE ***"
+  sudo apt update --allow-releaseinfo-change || true
+  if [ $? -ne 0 ]; then echo "ERROR: 'apt-get update' failed with error code: $?"; fi
   echo "*** INSTALLING REQUIRED BINARIES ***"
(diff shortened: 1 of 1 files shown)


setup.sh
```diff
@@ -10,14 +10,10 @@ if [[ $(cat /etc/os-release | grep -i debian) != "" ]]; then
   echo "Client is a Debian-based system. Installing binaries"; 
   echo
   echo "*** RUNNING APT-GET UPDATE ***"
-  sudo apt-get update --allow-releaseinfo-change
-  if [ $? -ne 0 ]; then echo "ERROR: 'apt-get update' failed with error code: $?"; exit 1; fi
-
-  echo
+  sudo apt update --allow-releaseinfo-change || true
```
vicwomg, line 13: This will disable the error checking on apt update. If apt fails to update, then the installs may result in outdated packages. 

I feel like if there's an error running apt update, it should be addressed by the end user rather than ignoring it.

vicwomg: If you force "|| true" though, wont the $? return 0? If that's the case then line 14 will never throw the error. 

On the other point, suppose a bad repository is accessed before the valid pikaraoke dependent-repositories are. The script would error AND the install would occur with out of date repositories. I don't see a way to handle this cleanly.

vicwomg:
pikaraoke is no longer installed in this way

Issue #293, opened 2023-12-25
Support Dockerized installation of pikaraoke


honestlai:
Love this project, wondering if it can be altered to be hosted.

I have a server that I run lots of docker stuff on. It crossed my mind that this would be great to dockerize if the splash screen could run outside of the host maybe over the net. I would of course protect it w/ something like SSO or an auth page via nginx.

I attempted to run the splash page externally on my systems browser with no luck. I'm guessing the audio stack in the scripting is pushing directly from the back end through the hosts sound system.

If this could be altered to have the audio+video render in the browser instead so it could run externally, that would be amazing!

-Matt

honestlai:
Hmm,

Now that I've read further into the documentation, I realized it sounds like --headless should accomplish what I'm trying to do, but when I que a song, my browser shows it's about to play, the title appears, but the video never start.

The logs show the song starting and then ending 10 seconds later... and then Song ending: none a minute or so after that. If I understand correctly, rendering should be done via ffmpeg and in the browser, but it's not streaming for some reason.

Anyone know why? 

I've currently got an instance spun up in an ubuntu docker routed through nginx on my work server to my home Mac systems browser.

vicwomg:
PiKaraoke is by default on port 5555 and Video streams over port 5556. So
make sure both are allowed through your firewalls. These ports are
configurable too, have a look at —help

Good luck! And do keep us updated if you get it working.

honestlai:
Digging in now to see about doing TCP forwarding using Argo tunnels. Is it possible to specify different streaming ports for the http and video streams?


Sent from Yahoo Mail for iPhone


On Monday, December 25, 2023, 3:49 PM, Vicwomg ***@***.***> wrote:



PiKaraoke is by default on port 5555 and Video streams over port 5556. So
make sure both are allowed through your firewalls. These ports are
configurable too, have a look at —help

Good luck! And do keep us updated if you get it working.

vicwomg:
Yes, there are CLI args for this:

```
  -p PORT, --port PORT  Desired http port (default: 5555)
  -f FFMPEG_PORT, --ffmpeg-port FFMPEG_PORT
                        Desired ffmpeg port. This is where video stream URLs will be pointed (default: 5556)
```

Run `pikaraoke.sh --help` for the full list of options

honestlai:
[...]

PR #301, opened 2023-12-28
Adding an ffmpeg url parameter

nathanpalmer:
This adds a way to override the ffmpeg_url. Specifically helpful in proxy situations where the website domain and the ffmpeg domain aren't going to be the same. It should fallback to the same URL it is using today if you don't pass in the parameter though.

Allows me to do this:

```bash
python3 app.py --url karaoke.example.com --ffmpeg-url karaoke-stream.example.com
```

The issue comes down to this. I don't want people who are coming to my house for the karaoke party to have to connect to my wifi network in order to use that QR code. I'd like to reduce as much friction as possible. So I have my existing network proxying the main URL going to the backend. However, we can't use that main URL as the ffmpeg_url because the PORT doesn't proxy through my router. (yes, I can setup a port forward.. but I had this dynamically assigning ports when using nomad so this option works a bit better.)

### DIFF
--- a/app.py
+++ b/app.py
@@ -831,7 +831,6 @@ if __name__ == "__main__":
         screensaver_timeout=args.screensaver_timeout,
         url=args.url,
         ffmpeg_url=args.ffmpeg_url,
-        prefer_ip=args.prefer_ip
         prefer_hostname=args.prefer_hostname
     )
 
--- a/app.py
+++ b/app.py
@@ -775,6 +775,13 @@ if __name__ == "__main__":
         required=False,
     ),
     parser.add_argument(
+        "-m",
+        "--ffmpeg-url",
+        help="Override the ffmpeg address with a supplied URL.",
+        default=None,
+        required=False,
+    ),
+    parser.add_argument(
         "--hide-overlay",
         action="store_true",
         help="Hide overlay that shows on top of video with pikaraoke QR code and IP",
@@ -833,6 +840,7 @@ if __name__ == "__main__":
         hide_overlay=args.hide_overlay,
         screensaver_timeout=args.screensaver_timeout,
         url=args.url,
+        ffmpeg_url=args.ffmpeg_url,
         prefer_ip=args.prefer_ip
     )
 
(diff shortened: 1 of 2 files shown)


vicwomg:
Good timing, we're discussing this in #293 which is an effort to dockerize everything. The same issue was encountered.

Can you explain a little more how you have things setup on your proxy?

vicwomg:
On second thought, can you try this branch and see if it suits your needs?

303-proxy_http_ffmpeg_streams

This will have pikaraoke do all the proxying to ffmpeg, so it will all run on one port.

nathanpalmer:
> Can you explain a little more how you have things setup on your proxy?

Sure. My router (Arista) is forwarding port 80/443 to a service router (Fabio) that handles the HTTPS termination and routes traffic to the right IP and port that it's running on. Since by default it uses the same url for ffmpeg and just appends the port, that doesn't make it through my router.

However, the other branch.. if it's all working on the same port sounds like it should also solve this issue. So I can try it.

vicwomg:
> > Can you explain a little more how you have things setup on your proxy?
> 
> Sure. My router (Arista) is forwarding port 80/443 to a service router (Fabio) that handles the HTTPS termination and routes traffic to the right IP and port that it's running on. Since by default it uses the same url for ffmpeg and just appends the port, that doesn't make it through my router.
> 
> However, the other branch.. if it's all working on the same port sounds like it should also solve this issue. So I can try it.

Yes please try the other branch and let me know. I believe that would be a cleaner solution overall

nathanpalmer:
> Yes please try the other branch and let me know. I believe that would be a cleaner solution overall

Sorry for the delay! We had a new years eve party and I didn't want to change anything right before that. But finally got a chance to try it out.

I opened up the other branch and it seems to work just fine as well. So either way that you want to go seems to be OK. I see the other branch is still open though. But either way will solve my problem.

I see there are some conflicts here, so if you want to go this direction I can resolve them. Otherwise, if you're going to merge the other branch in, we can close.

vicwomg (approved):
I'm going to go ahead and merge this. I don't see the harm in having it in and it seems to help with some folks with complicated networking needs

PR #305, opened 2023-12-28
Containerized branch based on 303-proxy_http_ffmpeg_streams

### LINKED ISSUE #293: Support Dockerized installation of pikaraoke

honestlai:
My first pull request. Not sure if I'm doing this right, but between GPT and KB's, things are happening. Resolves #293 

Future commits with a more detailed readme on the docker deployment scenarios and options coming soon.


honestlai:
In the master, I included my css changes to splash.html and  the full screen on button press when the splash page first loads

docker/entrypoint.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+
+# Correcting the URL format
+URL=$(echo "$URL" | sed 's|/$||')  # Remove trailing slash if exists
+if ! echo "$URL" | grep -q "^https://"; then
+  URL="https://$URL"  # Add https:// if not present
+fi
```
vicwomg, line 7: Is it possible a user might want to NOT run https?

docker/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Installation and Launch
+
+
+## Using Docker Compose Command Line
+
+1. **Clone the Repository**:
+   First, clone the repository from GitHub to get the `docker-compose.yml` file and any other necessary configurations:
+   ```bash
+   git clone https://github.com/honestlai/pikaraoke-docker.git
```
vicwomg, line 50: Would this URL actually be https://github.com/vicwomg/pikaraoke.git now?

vicwomg: Similarly line 51 should be `cd pikaraoke/docker` no?

docker/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+---
+
+## Support the Original Project
+
+If you want to support the original PiKaraoke project with a monetary tip, it's much appreciated by the original developer:
+
+[![Buy Me A Coffee](https://www.buymeacoffee.com/assets/img/custom_images/orange_img.png)](https://www.buymeacoffee.com/vicwomg)
+
+---
```
vicwomg, line 33: You can remove these lines from the docs since they are in the root-leve README

docker/README.md
```diff
@@ -0,0 +1,127 @@
...
+1. **Access Portainer**:
+   Open Portainer and navigate to the 'Stacks' section in the left sidebar.
+
+2. **Add a New Stack**:
+   Click on '+ Add stack'. Enter a name for the stack in the 'Name' field.
+
+3. **Compose File**:
+   Clone the repository from GitHub or copy the contents of the `docker-compose.yml` file into the 'Web editor'. The repository can be cloned using:
+   ```bash
+   git clone https://github.com/honestlai/pikaraoke-docker.git
```
vicwomg, line 71: As above, should this be https://github.com/vicwomg/pikaraoke.git ?

docker/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+5. **Deploy the Stack**:
+   After setting up your compose file and environment variables, click on 'Deploy the stack'. Portainer will pull the image from Docker Hub and start the container based on your configurations.
+
+## Building the container locally
+
+1. **Clone this Dockerized repository:**
+
+   ```bash
+   git clone https://github.com/honestlai/pikaraoke-docker.git
```
vicwomg, line 87: As above, should this be https://github.com/vicwomg/pikaraoke.git ?

templates/splash.html
```diff
@@ -265,6 +265,27 @@
       }
     });
   });
+
+  // Script for forcing browser to go full screen
+  function handleConfirmation() {
+    $("#permissions-modal").removeClass("is-active");
+    confirmationDismissed = true;
+    goFullScreen();
```
vicwomg, line 273: Personally, I could see this behavior of my browser going full screen to be unexpected. I personally prefer to work in windowed mode during active development. Another situation is if I'm in a multi-monitor setup and I wanted to drag the window to another 

Maybe make function take a `fullscreen: boolean` argument and add a separate button below confirm? e.g. "Confirm and go full screen"

templates/splash.html
vicwomg, line 1: Seems to be missing a couple of changes from the other branch that would fix safari / ios. Maybe we should merge that first, then merge master into this branch

docker/README.md
```diff
@@ -0,0 +1,127 @@
...
+3. **Accessing PiKaraoke:**
+
+   After the container is up and running, PiKaraoke will be accessible at `http://localhost:8888`. You can connect to this address from any device on the same network to access the PiKaraoke web interface.
+
+### Customizing Your Setup
+
+- Configuration options can be adjusted through environment variables in the `docker-compose.yml` file.
+- Songs are stored within a docker volume, and you can manage this storage as per your requirements.
+
+---
```
vicwomg, line 108: Same goes with lines 110+. No need for these in this doc

docker/entrypoint.sh
honestlai: Ahh... you are correct, they might want http as well. What I found was if I didn't include this, streaming didn't work properly. Ideally I'd like to leave this part out and get streaming functioning regardless if the user enters http, https, or just the hostname or fqdn.

Going to remove this portion for now.

docker/README.md
honestlai: Nice catch... I was too tired to update the readme last night. There's actually quite of a bit I'm going to add/change that should make deployment options more clear.

templates/splash.html
honestlai: Ya, I was having to do a little bit of musical chairs pulling down the right code that had some of these fixes. Also discovered a weird bug I'll report in the thread for that as well.

templates/splash.html
honestlai: Hmm.... ya, I see your point. I was really just trying to consolidate functions since a button press was already required. I've been hitting escape to get out of full-screen mode for dev work. I like your suggestion though.

templates/splash.html
honestlai: For now, I've set the Dockerfile to pull down from my branch. Once the ffnpeg fix and maybe some of my contributions make it into the main repo, I'll adjust the Dockerfile to pull from it.

docker/Dockerfile
```diff
@@ -0,0 +1,20 @@
+# Use Alpine as the base image
+FROM alpine:latest
+
+# Install required packages
+RUN apk add --no-cache bash zsh git gcc python3-dev musl-dev linux-headers py3-pip ffmpeg htop screen figlet
+
+# Clone the pikaraoke repository
+RUN git clone -b PiKaraoke-Docker https://github.com/honestlai/pikaraoke.git /pikaraoke
```
vicwomg, line 8: Does this need to clone a github repo at all? Could it simply copy the contents of ../ ? That way it is always buidling the image from the current code.

docker/entrypoint.sh
```diff
@@ -0,0 +1,22 @@
...
+
+# Exporting the corrected URL
+export URL
+
+# Run pikaraoke with environment variables
+if [ -z "$PASSWORD" ]; then
+  figlet PiKaraoke
+  /pikaraoke/pikaraoke.sh -d /pikaraoke-songs/ --headless -u $URL
```
vicwomg, line 15: Is it possible for $URL to be optional? I was trying this out and having issues trying to get a basic localhost instance running. It wants a URL, but inputting http://localhost:5555 seems redundant since the omission of -u would do that

templates/splash.html
```diff
@@ -408,11 +461,20 @@
...
   #qr-code {
     max-width: 25%;
+    border-top-right-radius: 20px; /* Rounded corner */
+    text-align: center; /* Center QR code */   
+    padding: 20px;
   }
   #up-next {
     max-width: 75%;
+    border-top-left-radius: 20px; /* Rounded corner */ 
+    padding: 10px 20px 10px 20px;
```
vicwomg, line 477: I had a look and I'm not crazy about these boxes. They cover up much more screen real estate than the text alone, most of it being blank space. If the goal is to make the text more readable, I feel that adding a black stroke effect to them would achieve the same thing and minimize how much of the video this is all covering up.  https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-stroke

templates/splash.html
```diff
@@ -408,11 +461,20 @@
...
     align-items: flex-end;
   }
+  #now-playing {
+    padding: 10px 20px 10px 20px;
+    border-bottom-left-radius: 20px;  
+  }
   #qr-code {
     max-width: 25%;
+    border-top-right-radius: 20px; /* Rounded corner */
+    text-align: center; /* Center QR code */
```
vicwomg, line 471: I'd prefer for this to remain right-aligned so it's less likely to cover karaoke video text.

templates/splash.html
vicwomg: On second thought, stroke is not great because it's an internal, not external stroke so it makes the text thinner. But an 8-axis shadow should work:
https://stackoverflow.com/a/47511171/2909171

docker/entrypoint.sh
```diff
@@ -0,0 +1,22 @@
...
+if ! echo "$URL" | grep -q "^https://"; then
+  URL="https://$URL"  # Add https:// if not present
+fi
+
+# Exporting the corrected URL
+export URL
+
+# Run pikaraoke with environment variables
+if [ -z "$PASSWORD" ]; then
+  figlet PiKaraoke
```
vicwomg, line 14: Minor nitpick, but  figlet is being run in both conditions, so you can remove it and place one just above the if block

docker/entrypoint.sh
honestlai: Removed... what would be ideal, but I couldn't really figure out, is for flask to determine what the endpoint is requesting in terms of port and address. I know the URL or hostname that's entered is self referential and makes its way into how the gui routes it's interface.

NGinx can easily send this data along back to the host... but the re-write of how to make this work on the back end was beyond me.

templates/splash.html
honestlai: A full screen button has been implemented, marking as resolved.

docker/Dockerfile
vicwomg: I messed around with this and I think it can work, but you need to .dockerignore the .venv directory

honestlai: It does, it's how docker containers stay up to date. In a production state the containers are generally sent to docker hub (much liike github), where there are versions that hold the contents of specific versions and base dependencies all within a mostly platform agnostic portable environment.

When a new build comes out from https://github.com/vicwomg/pikaraoke or my branch for instance, than I would build a new container, tag it with whatever syntax I'm using (still learning how proper container version syntax works), and update it on docker hub.

This way, for end users or sysadmins. If they find the :latest build that comes out, has some weird bug, or isn't compatible with an SQL plugin they're using (like my boss recently), the pull for the production environment can go backwards to something like vicwomg/pikaraoke-docker:1.12 on docker hub to satisfy what was working previously, reseting the ephemeral application layer of things, while the volume(s) (the users data and configs), remain and are unaffected by the downgrading of versions.

honestlai: Apologies if you know this already... making some assumptions here

docker/entrypoint.sh
honestlai: It can be, but with the way docker environments work, it will make the QR code function break. Docker environments network wise are like little subnetworks that sit beneath the network the docker host is sitting on. The spun up containers are almost like VM's because they all have their own IP within the docker subnet, and each container can be exposing their services on the same port without conflicting with one another because they are on their own individual ip.

By default, most containers will assign a randomized string of letters or numbers to their hostname, so without the URL flag, the QR code will read out something like ab362c89:5555 which will be completely unreachable to the end users. 

If you wanted to have the end users reach the docker host at 5555 on a local network, that would work, but I feel like most people who will be spinning up a containerized version of this project like me know that they will be using a reverse proxy to route an internet domain, or localized fqdn to the containers services.

docker/Dockerfile
vicwomg: I'm new to docker so this is all very educational and def. makes sense. Thanks for the crash course. I'll mull this all over

docker/entrypoint.sh
vicwomg: Super helpful info. I noticed it was putting in a blank or otherwise invalid hostname. Yes, I think if you're running locally, not using Docker would be simpler overall.

honestlai: What I might do is edit the coding so URL can be left blank, and if it is, the bash scripting will determine the docker hosts hostname and input this data for the user... so local hosting can still occur.

The resulting QR code would then read dockerhostname:5555 or dockerhostname.domain.local:5555

templates/splash.html
honestlai: Hmm.... I really like the look, to me it makes the interface very polished looking, and on most my screens it's rare they cover up important words. 

I think what I might try to do is configure this so by default it's off, but ad a check box in the admin area so it can be turned on and off dynamically.

templates/splash.html
honestlai: can do, not a biggie. It looked a little awkward when I had a long url there... I have one instance spun up on my personal server, and I know it would look odd to have the QR code left aligned.... buuuut even centered, the long URL + the QR doesn't look the greatest graphically either.

Been pondering a way to make it look more visually appealing... or maybe make the url text disappear altogether once a certain character count is reached.

vicwomg: Yeah, I agree its not great either way, maybe lowering the text size of the url would make it look more symmetrical. Or it might be worth adding a command line option to hide the url, but show the QR code, for minimalists. If that were the case a corner-aligned QR code would look pretty clean. But that's another ticket

templates/splash.html
vicwomg: I do like the general look, I think it's just the excessive blank space that bugs me. Maybe if the singer name goes on the same line as the song title, it would be cleaner. Then give it a max-width so it will wrap if necessary.

vicwomg: btw, I don't think it's trivial to add admin settings checkboxes that will affect client side frontend. You'd probably have to dip into flask

vicwomg: Another thought, maybe the project in general should allow css overrides for custom "skinning". There's no way any interface will make everyone happy.

docker/Dockerfile
honestlai: No worries... it's taken me years to wrap my head around docker... sometimes I feel like some kind of tech evangelist when I come across talented tech folk who have never used docker

"Have you heard of the good word from our lord and savior Docker?" 😅

I'm kubernetes and ansible folks feel the same way.

honestlai: Going to go ahead and mark this resolved

templates/splash.html
honestlai: Ya, I'm definitely feeling that with the QR code area at the moment.
image

docker/entrypoint.sh
honestlai: Oh! haha... not sure how I missed that.

templates/splash.html
honestlai: image

This might be worth adding as a tick box option in the admin area.

vicwomg:
ver 1.8.0 has full docker support over a single port

PR #318, opened 2024-02-16
Add command line arguments to specify width and height of the splash page

wesleong:
Thanks for sharing your code!

I've added a new command line parameter to specify the width and height of the chrome window that is opened. This was useful for me as I only installed the basic Xwindows without a window manager, so the --start-maximized flag doesn't do anything and I had to explicitly specify the window size.


vicwomg (approved):
Thanks! Looks great

PR #333, opened 2024-05-15
Add host AP info to new splash UI

wesleong:
This pull request adds the information about the Wifi AP name and password to the splash screen when using RaspiWifi, similar to how it looked before the switch to the web-based interface. In this instance, it displays a message that looks like the following in the upper left of the screen and also on the screensaver tile.

> Wifi Network: wifiName Password: wifiPassword
> Configure Wifi: http://10.0.0.1

This was tested on bullseye lite. For bookworm the wifi configuration is handled by networkmanager which breaks RaspiWifi functionality, so moving forward we will have to revisit this.

### DIFF
--- a/app.py
+++ b/app.py
@@ -41,6 +41,7 @@ app = Flask(__name__)
 app.secret_key = os.urandom(24)
 app.jinja_env.add_extension('jinja2.ext.i18n')
 app.config['BABEL_TRANSLATION_DIRECTORIES'] = 'translations'
+app.config['JSON_SORT_KEYS'] = False
 babel = Babel(app)
 site_name = "PiKaraoke"
 admin_password = None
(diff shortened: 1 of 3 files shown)


vicwomg:
Looks great, thank you for re-adding this feature. Good to know raspiwificonfig still, presumably, works. Is it possible to get a a screenshot of what the screen looks like? I haven't had the time to configure a pi to be in this state.

wesleong:
Sure! I'm not up to date with web interfaces so on the screensaver I just made the width larger, not sure how well it copes if the line of text is longer than the box though.

Regarding RaspiWifi, it looks like after Bullseye it will not work anymore unless someone updates that project for bookworm, which switched to using network manager to handle networking. I had to revert to bullseye for this reason. There is another project ([Comitup](https://github.com/davesteele/comitup)) that does work with bookworm, but probably requires some code changes in pikaraoke to support. For instance, it does not require a reboot to connect to a new network, so either we'd need to force a reboot or have pikaraoke use the loopback IP internally so that it doesn't break when it gets a new IP address.

Anyways, here is the main screen: 
image
And here is the screensaver view:
image

PR #391, opened 2024-09-03
Smoke Test

mariugul:
A small smoke test that pikaraoke can actually install and run `pikaraoke --help`. It works as a small sanity check that it didn't break and that it installs on all OS'es.

It works on Windows, macOs and Ubuntu :tada:


vicwomg:
Looks good. I guess I'll see how this affects my github billing :gulp:

vicwomg (approved):

mariugul:
> Looks good. I guess I'll see how this affects my github billing :gulp:

I am pretty sure it's free :open_mouth: Public repos should be free for this stuff. Let me know if it isn't :money_with_wings:

PR #405, opened 2024-10-11
Update get_platform.py

Carlos-GSan:
I installed pikaraoke on an OrangePi 3lts with Armbian SO and only works by adding this line of code

### DIFF
--- a/pikaraoke/lib/get_platform.py
+++ b/pikaraoke/lib/get_platform.py
@@ -57,6 +57,10 @@ def supports_hardware_h264_encoding():
     if is_raspberry_pi():
         platform = get_platform()
 
+        # For other platform(OrangePI etc)
+        if platform is None:
+            return False
+
         # Raspberry Pi >= 5 no longer has hardware GPU decoding
         match = re.search(r"Raspberry Pi (\d+)", platform)
         if match:


pikaraoke/lib/get_platform.py
```diff
@@ -57,6 +57,10 @@ def supports_hardware_h264_encoding():
     if is_raspberry_pi():
         platform = get_platform()
 
+        # For other platform(OrangePI etc)
+        if platform is None:
+            return False
```
vicwomg, line 62: Hmm, I feel like the real bug is that is_raspberry_pi() returns true for OrangePi. Is there a way to adjust is_raspberry_pi function to return false in the case of these devices?

vicwomg: More specifically, Open a python interpreter and run the following:

```
import sys
print(sys.platform)
```

What does the orangepi report as the sys.platform? If it's something unique from raspberry pi, I would add that condition to line 28 in the is_raspberry_pi function.

Carlos-GSan: orange@orangepi3-lts:~$ python3
Python 3.12.3 (main, Jul 31 2024, 17:43:48) [GCC 13.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import sys
>>> print(sys.platform)
linux
>>>

PR #423, opened 2024-11-22
feat: Added support to Android via Termux App

jepes1981:
This PR adds support to Android via the Termux App.

- lib/app.py 
Handle error when querying cpu usage on Termux (restricted)
Force to run on headless mode when detected running on Android.
- karaoke.py
Added way to query ip address when running on Termux/Android since sockets will not work.

- get_platform.py
is_raspberry_pi(): Added validation to avoid incorrectly reporting Termux/Android as raspberry pi.
get_platform(): Properly returns Android as platform to avoid crashes in other program logic.
- scripts/README.md
Added Termux/Android installation.

### DIFF
--- a/pikaraoke/app.py
+++ b/pikaraoke/app.py
@@ -553,7 +553,7 @@ def info():
     try:
         cpu = str(psutil.cpu_percent()) + "%"
     except:
-        cpu = "Android CPU usage query unsupported."
+        cpu = "CPU usage query unsupported"
 
     # mem
     memory = psutil.virtual_memory()
@@ -929,7 +929,7 @@ def main():
     cherrypy.engine.start()
 
     # force headless mode when on Android
-    if "Android" in platform and not args.hide_splash_screen:
+    if (platform == "android") and not args.hide_splash_screen:
         args.hide_splash_screen = True
         logging.info("Forced to run headless mode in Android")
     # Start the splash screen using selenium
--- a/pikaraoke/app.py
+++ b/pikaraoke/app.py
@@ -550,7 +550,10 @@ def info():
     url = k.url
 
     # cpu
-    cpu = str(psutil.cpu_percent()) + "%"
+    try:
+        cpu = str(psutil.cpu_percent()) + "%"
+    except:
+        cpu = "Android CPU usage query unsupported."
 
     # mem
     memory = psutil.virtual_memory()
@@ -925,6 +928,10 @@ def main():
     )
     cherrypy.engine.start()
 
+    # force headless mode when on Android
+    if "Android" in platform and not args.hide_splash_screen:
+        args.hide_splash_screen = True
+        logging.info("Forced to run headless mode in Android")
     # Start the splash screen using selenium
     if not args.hide_splash_screen:
         if raspberry_pi:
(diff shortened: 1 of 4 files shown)


scripts/README.md
```diff
@@ -44,6 +44,27 @@ setup-windows.bat
...
+```
+
+Install python and ffmpeg:
+
+```
+pkg install python ffmpeg
+```
+
+from this point onwards, you only need to follow the installation for Raspberry pi / Linux / OSX (see above).
+
```
vicwomg, line 67: I'd prefer to put this in the wiki and not the general readme

pikaraoke/lib/get_platform.py
```diff
@@ -24,15 +24,25 @@ def get_ffmpeg_version():
...
     try:
         return (
-            os.uname()[4][:3] == "arm" or os.uname()[4] == "aarch64"
-        ) and sys.platform != "darwin"
+            (os.uname()[4][:3] == "arm" or os.uname()[4] == "aarch64")
+            and sys.platform != "darwin"
+            and not (os.path.exists("/system/app/") and os.path.exists("/system/priv-app"))
+        )
     except AttributeError:
         return False
```
vicwomg, line 33: How about wrapping this in a function so we don't duplicate this code in the file:

```
def is_android():
  return os.path.exists("/system/app/") and os.path.exists("/system/priv-app")
```

pikaraoke/lib/get_platform.py
```diff
@@ -24,15 +24,25 @@ def get_ffmpeg_version():
...
     if sys.platform == "darwin":
         return "osx"
+    # elif sys.platform.startswith("linux"):
+    #    for key in os.environ:
+    #        if key == "PREFIX":
+    #            if "termux" in os.environ[key]:
+    #                return "Termux on Android"
+    #    return "linux"
+    elif os.path.exists("/system/app/") and os.path.exists("/system/priv-app"):
+        return "Android"
```
vicwomg, line 45: The convention is lowercase: `android`

pikaraoke/app.py
```diff
@@ -550,7 +550,10 @@ def info():
     url = k.url
 
     # cpu
-    cpu = str(psutil.cpu_percent()) + "%"
+    try:
+        cpu = str(psutil.cpu_percent()) + "%"
+    except:
+        cpu = "Android CPU usage query unsupported."
```
vicwomg, line 556: This might not be specific to android, `CPU usage query unsupported` will be more universal

pikaraoke/karaoke.py
```diff
@@ -189,18 +189,30 @@ def __init__(
...
-        try:
-            # doesn't even have to be reachable
-            s.connect(("10.255.255.255", 1))
-            IP = s.getsockname()[0]
-        except Exception:
-            IP = "127.0.0.1"
-        finally:
-            s.close()
+        # python socket.connect will not work on android, access denied. Workaround: use ifconfig which is installed to termux by default, iirc.
+        if "Android" in self.platform:
```
vicwomg, line 194: why not this more precise statement: `if (self.platform = "Android")`  (or `if (self.platform = "android")` if you follow my comment on making this lowercase)

vicwomg:
Great! Just had a few suggested changes. I will add the Android guide to the wiki

jepes1981:
I've addressed the requested changes. Please review them when you have time. Thank you!

PR #436, opened 2024-12-11
feat: on pi, possibility to change the audio output as well as use bl…

lvmasterrj:
I've created the possibility for a pi user, without access to the user interface, to change the oudio output from hdmi, headphone and bluetooth, as well as the possibility to scan and pair to bluetooth devices.

For it to work, it need the PipeWire with Wireplumber

`sudo apt install pipewire pipewire-pulse pipewire-jack pipewire-alsa pipewire-audio -y`

I created two files in the lib folder `audio_devices.py` and `bluetooth.py`.

- `audio devices.py` is the file responsible for getting the possible output audio devices, to connect to them and to increase and decrease their volumes. It uses `wpctl` from wireplumber.

- `bluetooth.py` is the file that manages bluetooth devices. It can scan for bt devices, connect and remove. It uses `bluetoothctl` to do the job. When the script pair to a devices it stores the device in a file called `btdevices.txt` inside the lib folder. Those are the devices that the system will automatically connect to when it finds it.

The user interface to work with these scripts are in the `info.html`.

I also created a function inside the `get_platform.py` to check if the pipewire is installed, if not it doesn't run these scripts.

One more thing: mainly because of the bluetooth scan, that takes some time, I've made it assyncronous so that the info page loads rigth away while this part is still loading and is updated when finished.

### DIFF
--- a/pikaraoke/app.py
+++ b/pikaraoke/app.py
@@ -22,6 +22,7 @@ from flask import (
     request,
     send_file,
     url_for,
+    jsonify,
 )
 from flask_babel import Babel
 from flask_paginate import Pagination, get_page_parameter
@@ -36,7 +37,7 @@ from selenium.webdriver.support.ui import WebDriverWait
 
 from pikaraoke import VERSION, karaoke
 from pikaraoke.constants import LANGUAGES
-from pikaraoke.lib.get_platform import get_platform, is_raspberry_pi
+from pikaraoke.lib.get_platform import get_platform, is_raspberry_pi, is_pipewire_installed
 
 try:
     from urllib.parse import quote, unquote
@@ -56,7 +57,6 @@ site_name = "PiKaraoke"
 admin_password = None
 raspberry_pi = is_raspberry_pi()
 linux = get_platform() == "linux"
-
 
 def filename_from_path(file_path, remove_youtube_id=True):
     rc = os.path.basename(file_path)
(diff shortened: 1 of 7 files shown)


vicwomg:
Apologies on the delay of looking on this PR. Was hard at work on 1.8 and didn't want to think about how this might conflict. I have a couple of concerns:

- This is very pi-specific and I'm trying to do less platform-dependent code
- This requires more binaries to be installed, which makes installation and documentation more complicated
- karaoke.py and app.py are becoming unsustainably huge, I'm thinking we should isolate utilities like this into their own classes and have their functions self-contained there. That way karaoke.py doesn't get cluttered up with more one-off functions.
- The info screen is getting very complicated. It would be good to modularize these components into separate files

I think I could be open to merging if we can figure out how to solve these issues

pikaraoke/app.py
```diff
@@ -707,6 +716,85 @@ def get_default_dl_dir(platform):
             return "~/pikaraoke-songs"
 
 
+@app.route("/audio_devices/")
+def get_audio_devices():
+    audio_devices = k.get_audio_devices()
+    bluetooth_devices = k.get_bluetooth_devices()
```
vicwomg, line 722: Do these need to be wrapped behind the k object? Can you just call these funcitons directly from the audio_devices lib?

PR #453, opened 2024-12-28
Release candidate 1.8.0

vicwomg:
Todos:

- [x] README update - remove venv stuff, add docker instructions
- [x] Deprecate unused scripts and libraries
- [ ] ~Fix tmpdir handling, be less aggressive about deleting files~
- [x] Translation audit
- [x] Support multiple background mp3 files, ~shoutcast streams~
- [x] Add more useful args to the info prefs page
- [ ] ~Explore need for the mov flags `frag_keyframe+default_base_moof` seem to break firefox~
- [x] Bug: song replays if queue is added during score screen
- [x] Smoke test on windows


Mcforwin:
Running on a single port and compatible with docker was just awesome.

I did had to do a couple of trial and error to deploy this in a TRUENAS scale running iX apps.
Could do it via yaml deploy or via the custom iX apps method which the later allows future modifications to the docker deploymentation.

The only thing that I had to change to make this work with iX apps was to change the docker file environment to commands. Otherwise, I have successfully deploy it base on the screenshot images shown below.
image
image
image
image
image

Maybe someone will be able to benefit from this if they are deploying the app via NAS.

PR #489, opened 2025-02-01
feat: a/v sync fix

lvmasterrj:
Added the possibility for the admin to fix the A/V synchronization.

- CLI Parameter: --avsync (Use avsync (in seconds) if the audio and video streams are out of sync. (negative = advances audio | positive = delays audio))
- User preference: avsync

I haven't test it with CDG/MP3 files

Now I'll create a "how -to" on how the user can adjust the a/v sync.

### DIFF
--- a/pikaraoke/app.py
+++ b/pikaraoke/app.py
@@ -174,6 +174,7 @@ def main():
         bg_video_path=args.bg_video_path,
         disable_score=args.disable_score,
         limit_user_songs_by=args.limit_user_songs_by,
+        avsync=args.avsync,
         config_file_path=args.config_file_path,
     )
 
--- a/pikaraoke/karaoke.py
+++ b/pikaraoke/karaoke.py
@@ -108,6 +108,7 @@ class Karaoke:
         disable_bg_video=False,
         disable_score=False,
         limit_user_songs_by=0,
+        avsync=0,
         config_file_path="config.ini",
     ):
         logging.basicConfig(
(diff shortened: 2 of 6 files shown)


lvmasterrj:
This is the Wiki text. If you want I can create a discussion topic (don't know whats best / easier for you).

=================================================

When running Pikaraoke with the sound output going to a Bluetooth speaker, due to Bluetooth latency, the speaker audio may have a slight delay compared to the video.

Here's how to Test and Fix Audio/Video synchronization:

### 1. Testing a/v sync
First, lets find an ideal video for testing the A/V sync.

1. On the search page, check the `Advanced` checkbox and select `Include non-karaoke matches`.
2. Search for `av sync` and choose one the suggested videos to start the test.

example of av sync test video
_Exemple of an AV Sync Test Video_

Since A/V sync test videos can vary pay attention to how the one you selected works.

The general idea is simple: a click sound will be synchronized with a visual cue (such as a bouncing ball, a flashing light, etc.). Your task is to determine whether the click sound occurs **before** or **after** the visual cue.

If you notice that the audio and the video are out of sync, let's fixit.

### 2. Fixing A/V sync

1. In the `Info` page, go to `Server settings`
2. There, you'll find a number box where you can adjust the audio timing relative to the video.

Now is trial and error.

- If the audio **lags behind** the video, set a **negative** value
- if the audio **is ahead** of the video, set a **positive** value.

Once you've adjusted the value, go back to `Browse` and play the A/V sync test video again. 

If everything is now perfectly synchronized, you can delete the test video and enjoy your karaoke session. If not, repeat Step 2 until it's properly set.

That's it! Hope this helps make your karaoke party even better! 🎤🎶

Issue #233, opened 2023-01-30
Feature Request: add audio track toggle button 


joefoxing:
There are a lot of video karaoke with two audio tracks, one with vocal and one without vocal. It wouldn't be complete for a karaoke app not to have an audio track toggle button on the control. Please consider adding this button.

PR #496, opened 2025-02-11
Support for audio track switch.

Kissycat:
Thanks for the awesome project !

The id of backing track could be set in the info page, it is 1 by default.

### DIFF
--- a/pikaraoke/routes/controller.py
+++ b/pikaraoke/routes/controller.py
@@ -1,4 +1,5 @@
 import flask_babel
+import time
 from flask import Blueprint, redirect, request, url_for
 
 from pikaraoke.lib.current_app import broadcast_event, get_karaoke_instance
@@ -8,10 +9,17 @@ _ = flask_babel.gettext
 
 controller_bp = Blueprint("controller", __name__)
 
+def pawait(s=1):
+    k = get_karaoke_instance()
+    if k.is_paused:
+        pause()
+        time.sleep(s)
+
 
 @controller_bp.route("/skip")
 def skip():
     k = get_karaoke_instance()
+    pawait()
     broadcast_event("skip", "user command")
     k.skip()
     return redirect(url_for("home.home"))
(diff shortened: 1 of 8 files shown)


Kissycat:
[One](https://github.com/vicwomg/pikaraoke/issues/233) more close~

pikaraoke/routes/controller.py
```diff
@@ -8,10 +9,17 @@
 
 controller_bp = Blueprint("controller", __name__)
 
+def pawait(s=1):
```
Kissycat, line 12: 1s may not enough for some case or device, maybe 2s is better.

vicwomg: We should avoid `time.sleep` whenever possible. What is this trying to solve? Is there a way to use a callback to improve synchonization?

Kissycat: > What is this trying to solve? 
Not so sure, I guess is the delay of ending last ffmpeg process and trying to start next one. 

> We should avoid time.sleep whenever possible.
You are right, we should solve the bug at root, but I have not found function existed for detecting ffmpeg, and what all I thought yesterday is how to finish track switch. Since today I am a guy want to play karaoke rather than write function~

So edit or drop the bad code as you wish please.

Maybe we should let the feats 'semitone change' and 'track switch' could remember the time already played next time. I was tried yesterday, but I not sure how the timestamp get the value of current&duration.

Kissycat:
Well, maybe it is not right place, anyway share a simple app for Android (tv) pikaraoke users.
```
Feats:
1. DNS query on https;
2. Shortcuts for phone;
3. Volume control via hardware volume keys;
4. En/简中;
5. None;
6. None;
......
```

Apk for [Download](https://luna.lovegood.cyou/s/lRhP)

Kissycat:
> It's possible to publish pikaraoke as an AndroidTV app? That's a fantastic idea.

Yeah, it is much more like a shortcut, save & open web page of pikaraoke, and only new webview support volume control via keys.
Of course,  we can run pikaraoke in termux the same time, hhh.

PR #518, opened 2025-07-19
[UPDATE] spanish translation 2025

leemongchupao:
here is some update of the translation to spanish

[es_VE.zip](https://github.com/user-attachments/files/21326810/es_VE.zip)


leemongchupao:
hope you can add it to the repo! I manage to update all the source text to spanish as accurate as I could!

vicwomg:
thank you!

PR #526, opened 2025-08-02
[Feat] Wi-Fi Setup feature for remote device user

fan9704:
# Wi-Fi Setup feature for portable device

### Add new Wi-Fi button in nav bar

image

Here I use emoji text istead of icon image if anyone know how to replace please help me

#### Show your device current Wi-Fi status

image

####Show your device available hot spot to connect

image

#### Click Connect button and enter password to connect

image

Show connected

image


vicwomg:
Thank you for your contribution. A few thoughts:

I don't think this should be in the main title bar. It is not common to change your wifi. It should just be a link under the settings page. No need for an icon.

Pywifi is not compatible with osx and it crashes launch with `NotImplementedError`. This error should be handled more gracefully. and I feel as though this is most relevant to pi users. Can you put this behind a launch flag like `--enable-wifi-configuration` which defaults to false.

fchangcelis:
i think this should be a part of the setting menu, because sometimes you have to change your wifi network and share it with your friends, good for access point and public wifi broadcast

vicwomg:
I think I’m fine with throwing a “platform not supported screen” in the
case of the error.

On second thought, No need to hide it behind a flag, but I do strongly
believe this should NOT be in the title bar. It should be in the settings
menu behind the is_admin permission. You don’t want basic users changing
the WiFi, accidentally or otherwise.

fan9704:
@vicwomg I remove Wi-Fi option from nav bar and move to info.html

Before

image

After

image

Also support i18n

vicwomg:
merging into a branch to resolve the translation conflicts

fan9704:
> I have moved this with the conflict resolutions to: #531
> 
> I have added some code to disable the feature for osx. However, neither Ubuntu nor windows launches the basic smoke test:
> 
> Ubuntu: `FileNotFoundError: [Errno 2] No such file or directory: '/var/run/wpa_supplicant'`
> 
> Windows: `ModuleNotFoundError: No module named 'comtypes'`
> 
> The pywifi library seems very limited in support. What did you test it on? Raspberry Pi? If so this might be a raspberry-pi-only feature. I am uncertain how useful wifi configuration from pikaraoke would be in non-pi systems since most users can just use the OS network configs to change wifi.

Test  in Windows 11

vicwomg:
we cannot merge this due to lack of cross platform support and outdated libraries

PR #539, opened 2025-11-09
add russian translate

avlubimov:
Thanks for the great program!
I translated it into Russian.


vicwomg:
I have manually added the russian translation to the dev branch. Feel free to open a ticket and submit another PR for alpine support in docker, but I'd like to know what the benefits are over sllim.

PR #541, opened 2025-11-10
Add Deno to dockerfile setup, add default tag to yt-dlp update

cdytoby:
yt-dlp will soon require deno (or other JavaScript runtime). This PR adds the relevant package and setup script to Dockerfile so that the docker image contains those dependencies.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "pikaraoke"
-version = "1.14.0"
+version = "1.15.0"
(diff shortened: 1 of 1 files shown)


vicwomg:
very interesting, I will read more about this change at yt-dlp and consider the best course of action

cdytoby:
I updated yt-dlp and python version, also updated the CI files, take a look and see if it's ok for you. 

I'll leave the rest of the documentation to you.

PR #551, opened 2025-11-18
Add download queue feature

cdytoby:
This PR improves download logic, so that the video gets downloaded one after another, in a queue.

- Extracted notification part from Karaoke class to its own class, so that it can be used anywhere.
- class `YtDlpClient` is created in file `youtube_dl.py`, and put download related code into it. Existing methods are also moved into this class.
- Add a queue system to download. 
- Additionally, make logging using `sys.stdout` instead of default `sys.stderr`, the default behavior writes all logs as error and makes it difficult for me to debug things.

Purpose of this PR:
- When multiple people in a party send download command, they don't all get start download, so that bandwidth can focus on a single download at a time.
- Try to reduce the chance that certain error appears, which indicates host is a robot and require log in.

Side note:
- I used AI to assist me doing this. I use AI Assistant in PyCharm, and used Claude as LLM. I personally reviewed all the codes, and made many manual adjustments accordingly. 
- I know it takes time to review the big change, so no rush, take your time and review it properly.


cdytoby:
I rebased the branch and made some changes. This PR is now ready to review. 

However, I have trouble understanding what exactly fails in code-quality CI check. Can someone help me with this? CI already performs reformatting successfully, so I doubt this is the issue here.

vicwomg:
I've added: https://github.com/vicwomg/pikaraoke/pull/601 with an approach that seems to be a bit more like how I'd want this feature to work. Have a look and let me know if that implements download queue as expected

PR #572, opened 2025-12-18
Replace polling with event-driven architecture

safepay:
In the interests of efficiency, looking at the code, the 0.5 second polling can cause overhead to a RPi.

In this Pull Request:
Replace 500ms polling loop with direct SocketIO emissions on state changes. This reduces CPU overhead by ~75% on resource-constrained devices.

Changes:
- karaoke.py: Accept socketio parameter, emit directly on state changes
- karaoke.py: Remove MD5 hash calculations (only used for polling detection)
- app.py: Remove poll_karaoke_state() and polling thread
- app.py: Pass socketio instance to Karaoke constructor

Technical details:
- Kept method names (update_now_playing_hash/update_queue_hash) for minimal diff
- Methods now emit via socketio instead of computing MD5 hashes
- Notifications emit immediately when sent (vs. next poll cycle)
- No client-side changes required (same event names/payloads)

Result: State changes trigger immediate updates instead of waiting for poll cycle.

### DIFF
--- a/pikaraoke/templates/splash.html
+++ b/pikaraoke/templates/splash.html
@@ -7,7 +7,6 @@
   rel="stylesheet"
   href="{{  url_for('static', filename='score.css') }}"
 />
-<script src="https://cdn.jsdelivr.net/npm/hls.js@latest"></script>
 <script>
   const isMediaPlaying = (media) =>
     !!(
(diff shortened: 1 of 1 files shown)


safepay:
I updated the PR to only include one change and made a separate PR for the other.

safepay:
Closing as superceded.

PR #573, opened 2025-12-18
Optimize video streaming for smooth playback

safepay:
Reduce streaming latency by fixing sleep interval and chunk size. This eliminates stuttering during transcoded video playback.

Changes:
- Reduced sleep from 1000ms to 50ms (20x faster response)
- Reduced chunk size from 25MB to 2MB (less memory pressure on Pi)
- Removed redundant file.seek() and position tracking
- Simplified chunk reading logic
- Read all remaining data after transcoding completes

Impact:
- Eliminates 1-second gaps between chunks during playback
- Reduces memory footprint for streaming (critical on Pi Zero/1)
- Affects all MP3+CDG files and pitch-shifted videos (most karaoke use)

Technical notes:
- 50ms sleep only when no data available (yields CPU without blocking)
- Stream continues smoothly while ffmpeg transcodes in background
- Browser receives continuous chunks instead of stuttering every second

### DIFF
--- a/pikaraoke/routes/stream.py
+++ b/pikaraoke/routes/stream.py
@@ -20,20 +20,22 @@ def stream(id):
     k = get_karaoke_instance()
 
     def generate():
-        position = 0  # Initialize the position variable
-        chunk_size = 10240 * 1000 * 25  # Read file in up to 25MB chunks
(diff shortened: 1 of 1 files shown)


safepay:
Unfortunately I don't have a complex test setup to be able to test most of these changes. I don't want to waste your time, but this should work.

PR #580, opened 2025-12-23
Improve streaming to all devices, settings handling and several key bug fixes.

safepay:
Firstly, sincere apologies for a big pull request but lot's of overlapping changes and it was too hard to have separate pull requests for this.

This started in me trying to get this streaming to my Smart TV's, which failed due to the lack of fMP4 support, only HLS.

The key underlying change is that it now transcodes to m3u8 and then serves this either directly or in a fMP4 "wrapper". It waits for 2 chunks to be transcoded before starting the stream, which only takes a couple of seconds.

With that solved, I then fixed several other bugs, such as not skipping to the next song after pausing and not persisting the settings across restarts. Setting handling is much improved because there was previously a reliance on the strings "True" and "False" rather than the booleans. It is now all boolean based.

The main outcome is that you almost never need to fully transcode before playing. There is almost no delay in playing from more powerful hardware. Even streaming from the RPi is pretty quick. Kiosk mode on RPi is Ok, given the overhead from the desktop. If you could run in kiosk mode without the desktop, it would be even better.

Here are the changes:

- HLS and MP4 stream switching via hardware auto-detection and new --streaming-format flag (auto/hls/mp4)
- Auto-detection of streaming format based on hardware
- Fixed setting handling from strings to booleans.
- Enables encoding and streaming on the fly from RPi and Windows in both kiosk and headless modes
- HLS now supports Smart TVs with fMP4 segments and native HLS playback (fixes audio sync issues)
- Fixes settings persistence across restarts - CLI arguments properly override saved config, UI preferences persist correctly
- Fixes Pause, Skip, and Restart processes for both MP4 and HLS streams
- Event-driven architecture - replaced 500ms polling with direct SocketIO emissions (~75% CPU reduction on RPi)
- Improved audio/video sync with vsync parameter and constant frame rate for HLS
- Kiosk mode improvements - Chrome flags for better media control in Windows 10 kiosk mode

I have tested on Windows and RPI, in both head and headless modes, streaming from one to the other as well as to smart TV's from LG and Samsung.

For example, on a RPi, run "--headless --steaming format hls" and you can stream to a smart TV's browser.

I hope you are able to make use of these changes.

Merry Xmas!

### DIFF
--- a/pikaraoke/lib/ffmpeg.py
+++ b/pikaraoke/lib/ffmpeg.py
@@ -102,11 +102,8 @@ def build_ffmpeg_cmd(
     else:
         video = input.video
 
-        # For WEBM files with VFR, force constant 30fps to fix timing issues
-        if fr.file_extension == ".webm":
-            video = video.filter("fps", fps=30).filter("setpts", "PTS-STARTPTS")
-            # Resample audio and reset timestamps to fix timing alignment
-            audio = audio.filter("aresample", 48000).filter("asetpts", "PTS-STARTPTS")
+        # For WEBM files, genpts at input level handles timestamp regeneration
+        # No additional filters needed - let genpts + avoid_negative_ts do the work
 
         # Output format based on streaming_format setting
         if fr.streaming_format == "mp4":


safepay:
I know. I know. All those commits are a pain in the arse.
I'll push another commit that fixes the mov, mkv and avi playback issue. It's not too hard.

safepay:
Key changes were to resolve audio problems, plus a fix for corrupted video due to reuse of the init.mp4 wrapper.

Should be good to go. Working nicely for me.

PR #601, opened 2026-01-02
feat: serialize/queue downloads from yt-dlp

vicwomg:
Serialize downloads to minimize CPU, bot detection, and network strain.


cdytoby:
Thanks, I'll just try the release version and create issue if I found any.

PR #611, opened 2026-01-02
perf: reduce docker image size, misc improvements

safepay:
- Remove Poetry from runtime image (use pip directly with PEP 621 `pyproject.toml`)
- Add non-root user for security
- Pin all versions (Python 3.12.8, Deno v2.1.4) which improves reproducability
- Add pip optimization flags (`--prefer-binary`, `--no-compile`)
- Add health check, expose port, volume declarations
- Suppress debconf warnings with `DEBIAN_FRONTEND=noninteractive`
- Combine RUN commands to reduce layers

Add a .dockerignore file to reduce build size.

Overall this reduces the image size by about 20% due to the lack of Poetry.


safepay:
> Happy to merge this if you've tested it working (I don't have docker set up right now), I kind of hate Poetry and I'm surprised I didn't think of it, haha

You can't think of everything. It's a pretty comprehensive program to start with!

Issue #615, opened 2026-01-03
[FEATURE] Move the iFrame to a Single Page Architecture


safepay:
I'll raise this as a feature request before working on it because it represents a significant change.

The iFrame is legacy and given that PiKaraoke already uses Ajax, we should move the iFrame to a DIV and load the content dynamically. This would smooth the process for the user and take resource pressure from the browser.
The only caveat is that IE <9 would not be supported, but I think we can live with that!

Then in the future, as a bigger upgrade, we can migrate to Fetch.

Thoughts welcome.

safepay:
You can see progress on my branch here:
https://github.com/safepay/pikaraoke/tree/claude/iframe-to-div-migration-Rjfvk

Mainly CSS cleanup. I think there are overlapping css styles.

safepay:
I only have one bug to deal with which is handling queue clearing. The issue stems from the dialogue boxes which need to be handled slightly differently.

But the overall code is working well and brings a much smoother experience to the user.

Definitely worth testing at this point.

safepay:
Ok. I'm pretty happy with this branch. All full page refreshing is removed which improves the usability across devices. RPi should be snappier s less is happening in the browser.

vicwomg:
nevermind, looks correct on the PR. Not sure if I checked out the wrong branch on your fork.

PR #619, opened 2026-01-04
feat: optimize web UIs with SPA architecture, add --browse-results-per-page CLI parameter

safepay:
This PR addresses #615.

The pages used to fully refresh when the user navigated throughout the site. This was quite a pain on an old RPi especially, but also not a great overall user experience.

The uses the existing Ajax library and extends it beyond the functions that already existed to all of the navigation and dialogue buttons.

This gives a better mobile experience too and feels more like an app than a web page. I also added configurable results per page in the browse section which can be set to say 200 which makes the mobile and RPi interfaces pretty snappy. I hope you lieke this idea.

The iFrame is retained.

This also fixes type errors causing settings to not be retained while browsing.

### DIFF
--- a/pikaraoke/routes/admin.py
+++ b/pikaraoke/routes/admin.py
@@ -203,22 +203,33 @@ def auth():
         type: string
         required: true
         description: Admin password
+      - name: next
+        in: formData
+        type: string
+        required: false
+        description: URL to redirect to after successful authentication
     responses:
       302:
-        description: Redirects to home on success, login on failure
+        description: Redirects to next URL on success, login on failure
     """
     d = request.form.to_dict()
     admin_password = get_admin_password()
     p = d["admin-password"]
+    next_url = d.get("next", "/")
+
+    # Validate next_url to prevent open redirect vulnerabilities
+    if not next_url.startswith("/"):
+        next_url = "/"
+
     if p == admin_password:
-        resp = make_response(redirect("/"))
+        resp = make_response(redirect(next_url))
         expire_date = datetime.datetime.now()
         expire_date = expire_date + datetime.timedelta(days=90)
         resp.set_cookie("admin", admin_password, expires=expire_date)
         # MSG: Message shown after logging in as admin successfully
         flash(_("Admin mode granted!"), "is-success")
     else:
-        resp = make_response(redirect(url_for("login")))
+        resp = make_response(redirect(url_for("admin.login", next=next_url)))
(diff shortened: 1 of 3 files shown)


vicwomg:
Looking good. Still poking around but noticed a couple of things:

- Boolean settings are not persisting in the settings menu. In fact this affects the current master branch. Probably related to your previous changes around this area.
- Clicking the menu > pikaraoke to go to settings should close the menu (nitpick)

safepay:
> * Clicking the menu > pikaraoke to go to settings should close the menu (nitpick)

This is fixed but still working on the settings persistance.

And I checked the persistance in the current release and it works fine for me.

safepay:
Added "--browse-results-per-page" as a startup setting and user-configurable because with the new Ajax handling, a RPi is getting almost quick (!) and this would make it even faster.

Plus, depending on the number of songs in a library, a user can make the mobile interface more snappy.

pikaraoke/karaoke.py
```diff
@@ -388,8 +388,40 @@ def change_preferences(self, preference: str, val: Any) -> list[bool | str]:
...
+            if preference in FLOAT_PREFERENCES:
+                try:
+                    typed_val = float(val_str)
+                except ValueError:
+                    typed_val = val_str
+            # Check for integer first (e.g., "1", "0", "42", "-5")
+            elif val_str.lstrip('-').isdigit():
+                typed_val = int(val_str)
+            # Check for boolean strings (excluding "1" and "0" which are handled as numbers)
+            elif val_lower in ("true", "yes", "on"):
```
vicwomg: The same issue occurs on line 360 upon loading preferences. If a loaded config is 0, then it gets set to False. 

I have a PR fixing this. You may just want to revert this and pull that down once I merge.
https://github.com/vicwomg/pikaraoke/tree/fix-settings-persistence-broken

I think this bug is severe enough for me to want a fix out asap and not wait for the testing of this branch

safepay:
Ok. Ready for testing and merging.

vicwomg:
Great. Looking awesome so far from what I’ve seen. I’ll give it a pass
tonight

safepay:
I've enjoyed working on this with you. It's a great program and I just wanted to make it better. It was pretty showing my sister in law today. I ripped 600.somgs from her old discs for and set it up for her so she is just rapt.

This should be it for the big commits, but I still have a couple of ideas left!

PR #642, opened 2026-01-06
Add Feature: Manually triggered Windows executable build

safepay:
This PR puts all the files in place plus a Git action to allow manually triggering of a Windows install program for PiKaraoke.

The EXE allows any of the flags to be passed at startup, just like the python-installed version.

The caveat...

Users will see (The "Scary" Warning)

When someone downloads PiKaraoke-Setup.exe from GitHub and tries to run it, they will likely see the Microsoft Defender SmartScreen warning because the app is unsigned and "uncommon."

It looks like a blue error screen:
- Windows protected your PC Microsoft Defender SmartScreen prevented an unrecognized app from starting...

They will have to:
- Click "More info" (which is small text).
- Click the "Run anyway" button that appears.

The solution for now is to simply let users know that this will happen and why.

The next step would be to submit it to the MicroSoft store. This would be the cheapest way to get the program published and signed at the same time.

I have tested locally to show that the installer can run silently and successfully, just like an App installer. The caveat is that the pikaraoke-songs directory defaults to live in the user's Documents directory and ffmpeg is automatically bundled with the installer, rather than being a user choice. This should not be a problem.

It would be a manual process to upload the compiled EXE to the MS Store each time a release is created via the workflow.

### DIFF
--- a/pikaraoke/app.py
+++ b/pikaraoke/app.py
@@ -21,7 +21,7 @@ from pikaraoke.lib.args import parse_pikaraoke_args
 from pikaraoke.lib.current_app import get_karaoke_instance
 from pikaraoke.lib.ffmpeg import is_ffmpeg_installed
 from pikaraoke.lib.file_resolver import delete_tmp_dir
-from pikaraoke.lib.get_platform import get_platform, has_js_runtime
+from pikaraoke.lib.get_platform import get_data_directory, get_platform, has_js_runtime
 from pikaraoke.lib.selenium import launch_splash_screen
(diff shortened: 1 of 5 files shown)


vicwomg:
Very interesting indeed. Can we move this to some kind of subdirectory in the project? Prefer not to have it in the project root  something like /build_scripts/windows

vicwomg:
by the way, not sure if you accepted the collaborator invite but if so you can create branches/PRs directly in the master project now if you prefer.

safepay:
Closing this for now. There are too many issues still to resolve.

PR #643, opened 2026-01-06
feat: Refactor file storage to use platform-specific data directories

safepay:
Currently, the application attempts to write runtime files (QR codes, config files) to the application's root directory. This causes permission errors on strict environments (like Windows Program Files) and creates clutter.

Changes This PR moves all program-generated files to the appropriate user data directory for each platform (e.g., %APPDATA% on Windows, ~/.pikaraoke on Linux/Mac).
Specifically, config.ini and qrcode.png

Benefits
- Fixes Windows Permissions: Resolves PermissionError crashes when the app is installed in read-only directories.
- Improved Layout: Standardizes file organization across all operating systems.
- Future Proofing: Establishes the correct, writable location for the upcoming SQLite database integration.

### DIFF
--- a/tests/unit/test_get_platform.py
+++ b/tests/unit/test_get_platform.py
@@ -261,28 +261,33 @@ class TestGetDataDirectory:
     """Tests for the get_data_directory function."""
 
     def test_windows_path(self):
-        """Test that Windows returns the APPDATA path.
-
-        We mock os.path to be ntpath (Windows path logic) to ensure join
-        uses backslashes even if tests run on Linux.
-        """
+        """Test that Windows returns the APPDATA path."""
         with patch("pikaraoke.lib.get_platform.is_windows", return_value=True):
             with patch.dict(os.environ, {"APPDATA": "C:\\Users\\Test\\AppData\\Roaming"}):
-                with patch("os.path.exists", return_value=True):
-                    with patch("pikaraoke.lib.get_platform.os.path", ntpath):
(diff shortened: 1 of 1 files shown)


pikaraoke/constants.py
```diff
@@ -15,3 +18,24 @@
...
 }
+
+
+def get_data_directory():
+    """
+    Returns the writable data directory for the application.
+    Windows: %APPDATA%/pikaraoke
+    Linux/Mac: ~/.pikaraoke
+    """
+    if sys.platform == "win32":
```
vicwomg: we should use the existing get_platform() function

pikaraoke/constants.py
```diff
@@ -15,3 +18,24 @@
     "zh_Hans_CN": "Chinese (Simplified)",
     "zh_Hant_TW": "Chinese (Traditional)",
 }
+
+
+def get_data_directory():
```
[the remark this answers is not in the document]
safepay: Yes. That is a much better place for it. Not sure what I was thinking.

pikaraoke/constants.py
safepay: > we should use the existing get_platform() function

I thought it would be better as a separate function to go along side the get_dl, get_os, etc.

safepay:
Thanks. I've been a bit busy/tired to apologies for these oversights.

vicwomg (approved):

PR #645, opened 2026-01-08
 feat: Add Multi-Platform Standalone Binary Builds (Win, Mac, Linux)

safepay:
**Summary**
This PR enables the automated generation of standalone executables for all supported platforms (Windows, macOS, and Linux) using PyInstaller and GitHub Actions.

**Briefcase**
- Uses Briefcase to easily create cross-platform installers.
- Simple and maintainable configuration via pyproject.toml
- Does not optionally install ffmpeg on Windows, but this can be added.
- Can also support Android builds in the future to have a single build process


vicwomg:
miraculously, it seems to work if I disable flasggr and add the platform import to the spec. 

MacOS build is around 120MB. I'm a little puzzled by the structure, there is a `pikaraoke` executable and a `_internal` folder filled with python deps. It will not work without the `_internal` folder. Not sure how to distrubute something like that.

.github/workflows/release-pi-arm64.yml
```diff
@@ -0,0 +1,57 @@
...
+        image: python:3.10-slim-bookworm
+        # Mount the current repo to /work inside the container
+        options: --platform linux/arm64 -v ${{ github.workspace }}:/work
+        run: |
+          cd /work
+
+          # A. Install System Dependencies
+          echo "Installing system dependencies..."
+          apt-get update
+          apt-get install -y git build-essential ffmpeg
```
mariugul: This is also a typical maintenance black hole for the future. Find a better way

.github/workflows/build-all-binaries.yml
```diff
@@ -0,0 +1,111 @@
...
+    - name: Install Inno Setup
+      if: matrix.os == 'windows-latest'
+      run: choco install innosetup
+
+    - name: Build Windows
+      if: matrix.os == 'windows-latest'
+      # This script runs PyInstaller AND Inno Setup
+      run: powershell .\build_scripts\windows\build_windows_installer.ps1
+
+    # --- UNIX BUILD (Direct PyInstaller) ---
```
mariugul:
```suggestion

```
obvious comment

.github/workflows/release-pi-arm64.yml
```diff
@@ -0,0 +1,57 @@
...
+
+    steps:
+    - name: Checkout Code
+      uses: actions/checkout@v4
+
+    # 1. Enable emulation support for ARM64 (Raspberry Pi architecture)
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v3
+
+    # 2. Run the build inside a Docker container that matches Raspberry Pi OS (Debian Bookworm)
```
mariugul:
```suggestion

```

build_scripts/windows/build_windows_installer.ps1
mariugul: It's a medium-complicated script. Can it be simplified?

mariugul (dismissed):
It's really cool to have one binary for macOS, Linux and Windows. It's quite complicated. 800+ lines of code to support this. This will increase maintenance quite a bit. Any new dependency, tool or feature might break the whole chain. I would at least suggest this get scaled down to less code and simpler code.

.github/workflows/build-all-binaries.yml
```diff
@@ -0,0 +1,111 @@
...
+        os: [windows-latest, ubuntu-latest, macos-latest]
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.10'
+        cache: 'pip'
```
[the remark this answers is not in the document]
mariugul: `uv` handles python already. It will install it for you if lacking. Just specify to `uv` the lowest supported python version.

vicwomg:
> It's really cool to have one binary for macOS, Linux and Windows. It's quite complicated. 800+ lines of code to support this. This will increase maintenance quite a bit. Any new dependency, tool or feature might break the whole chain. I would at least suggest this get scaled down to less code and simpler code.

Yes quite complicated, and if there are optimizations that can be made then great. I do think it's worthwhile to explore binary builds and try for a time. 

Most people non-technical are terrified of the CLI. This breaks down that hurdle and expands the user base.

build_scripts/docker/docker-compose.yml.example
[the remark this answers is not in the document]
vicwomg: it's just an example file I added for anyone who needs a reference when creating their own (I personally forget how to use docker every time)

safepay:
> Or spend some time investigating whether there are other clever ways to subvert the browser security permissions around autoplay.

See this branch:
remove-selenium-and-confirm-dependencies
Still a couple of issues, but seems to be working.

vicwomg:
LMK if you are ok with closing this. I think with the one-line install script we're getting most of the benefits here without having to force a command line app into a standalone binary. Ultimately, I think we're always going to have some awkwardness given pikaraoke is really a CLI app and it has external binary dependencies, so it can't really be self contained.

PR #646, opened 2026-01-08
fix: flasgger incompatibilities, disable by default

vicwomg:
flasgger adds extra overhead and breaks the pyinstaller build, let's disable by default behind a CLI flag. As far as I can tell this just creates a flask endpoint to display the API. Not critical or likely needed by anyone except for devs.


mariugul (commented):
This is absolutely a respectable choice. I would suggest to consider another option also which would be to set and read a DEV var. For example if DEV env var is set. It would be more generic. Just a different way of doing it, not necessarily better.

PR #659, opened 2026-01-13
fix: slow info.html page load due to CPU check

vicwomg:
I noticed the info.html page hangs for a second or so on load, unlike other routes. Turned out to be the CPU check which is a blocking action. Adjust the CPU check to occur asynchronously

### DIFF
--- a/pikaraoke/routes/info.py
+++ b/pikaraoke/routes/info.py
@@ -2,7 +2,7 @@
 
 import flask_babel
 import psutil
-from flask import Blueprint, render_template
+from flask import Blueprint, jsonify, render_template
 
 from pikaraoke import VERSION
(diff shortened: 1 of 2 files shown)


safepay:
Actually. Try changing it to 0.1 seconds.

That way the reading will be more accurate but not block the page loading.

vicwomg:
> And trying to avoid sneaky commits ;-) maybe change the Chinese to English on line 39!

Sure thing! that's fine for insignificant changes like comments clean up.

> Actually. Try changing it to 0.1 seconds.
> That way the reading will be more accurate but not block the page loading.

I changed it back to 1 and added a new ajax route for this data so we can load it async and show a loading text instead. This is probably the right way.

PR #676, opened 2026-01-17
feat: drag+drop queue reordering, move to top/bottom

vicwomg:
- added drag and drop via touch/click and drag
- added modal menus to provide mulitple options
- add current playing song functions (restart, pause, skip) to modal menu


safepay:
Great stuff.

I think this plus the round-robin PR I just added really enhances queuing and queue management. Once these are merged I'll factor out the queuing functions into a serparate file as per the plan.

PR #677, opened 2026-01-17
feat: Smart Queue (Round-Robin Fair Play)

### LINKED ISSUE #132: Feature request, improved que/singer rotation management
When a singer adds a song to the que, it just goes to the bottom of the que. If singer 1 adds 3 songs, all 3 songs will be back to back, unless someone shuffles the que around to reflect the rotation of singers. It would be nice if singers could add as many songs as they want, and manage songs in their individual que, while the playback que pulled songs from the top of the singers que by singer rotation. So if singer 1 picks 3 songs, singer 2 put in 2 songs, singer 3 put 1 song in, singer 4 put 4 songs in, the rotation would automatically go singer 1, 2, 3, 4, 1, 2, 3, 4, 1, 2, 4, 4 etc. When singers added songs, it would update that singers que, and playback que draw songs from each singers que in order of singer rotation. Singer rotation would need a way to be managed. A separate html interface could exist for the KJ host to have a rotation manager tab, so they could adjust the rotation, and pause or delete singers in the playback que as needed. The existing interface would look the same, except singers would view and change their own singer rotation, and maybe view only the next song of all singers (or the next 3 or 5 or something like that) in the playback rotation. The host interface would have an extra tab for rotation management, to move pause or delete singers in the playback que. This might be a big ask, but if just the rotation que worked even without the rotation management, I think it would be a great improvement.

safepay:
Resolves #132 

Songs are now inserted using round-robin scheduling instead of simple append. This prevents one user from monopolizing the queue.

**How it works:** Each user takes turns. Your 2nd song goes after everyone else's 1st song, your 3rd song goes after everyone else's 2nd song, etc.

**Example:**
- Alice adds 3 songs, Bob adds 1 song
- Queue order: Alice-1, Bob-1, Alice-2, Alice-3
- Old behavior: Alice-1, Alice-2, Alice-3, Bob-1

### DIFF
--- a/pikaraoke/karaoke.py
+++ b/pikaraoke/karaoke.py
@@ -279,9 +279,7 @@ class Karaoke:
             else limit_user_songs_by
         )
         self.enable_fair_queue = (
-            pref
-            if (pref := self.get_user_preference("enable_fair_queue")) is not None
-            else False
+            pref if (pref := self.get_user_preference("enable_fair_queue")) is not None else False
         )
         self.cdg_pixel_scaling = (
             pref
--- a/pikaraoke/karaoke.py
+++ b/pikaraoke/karaoke.py
@@ -729,6 +729,42 @@ class Karaoke:
         )
         return True if cont >= int(self.limit_user_songs_by) else False
 
+    def _calculate_fair_queue_position(self, user: str) -> int:
+        """Calculate insertion position for round-robin fair queuing.
+
+        Implements Nagle Fair Queuing: users take turns in rounds. A user's Nth
+        song is placed after all other users' Nth songs (or at queue end).
+
+        Args:
+            user: Username adding the song.
+
+        Returns:
+            Queue index where the song should be inserted.
+        """
+        # Count how many songs this user already has in queue
+        user_song_count = sum(1 for item in self.queue if item["user"] == user)
+
+        # Find position after the last song in "round N" where N = user_song_count
+        # Round 0 = first song from each user, Round 1 = second song, etc.
+        target_round = user_song_count
+        songs_seen_per_user: dict[str, int] = {}
+
+        for idx, item in enumerate(self.queue):
+            queue_user = item["user"]
+            songs_seen_per_user[queue_user] = songs_seen_per_user.get(queue_user, 0) + 1
+            # This song is in round (count - 1) for its user
+            song_round = songs_seen_per_user[queue_user] - 1
+            if song_round == target_round:
+                # Found a song in the target round, insert after it
+                # Keep scanning to find the LAST song in this round
+                pass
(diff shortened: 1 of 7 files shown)


safepay:
> I read the article you'd linked by "This guy" and realised you'd found my long-lost twin (at least as far as his views on Karaoke etiquette are concerned). I can't comment on the technical implementation aspects of your merge request, but the intent is something I'd promote at every opportunity! There are well-known professional karaoke service providers whose interfaces lack this very same 'fairness' feature (or they lacked them the last time I'd encountered their crappy queue management a year or two back). Let fairness prevail!

Thanks for the feedback. The question is not whether this is included or but whether it should be the default behaviour with a fall back to current behaviour or vice versa. What is your opinion?

vicwomg:
My suggestion to make this non-default has nothing to do with whether I think it's useful. I think it is something I would absolutely turn on. 

However a chronological queue is the most intuitive default situation for a new user. I can forsee someone now familiar with the system adding a song and being puzzled as to why it isn't in the queue where they expect, then try to add it again and get an error.

As a feature toggle, the user is thus required to understand what they are enabling before they do it.

safepay:
No worries. Can I suggest that because you now have persisted settings, that there is less need to keep adding command line arguments for UI-things?

Other command line arguments that are REQUIRED for correct operation at start-up are obviously required, but elements such as --browse-results-per-page and, for example "--round-robin-queuing" are not required. They can just be settings in the admin page that get persisteed across restarts.

The persistance feature gives you a lot more freedom to add features without clogging up the command line!

I would actually recommend REMOVING the --browse-results-per-page if it's not too late.

vicwomg:
> No worries. Can I suggest that because you now have persisted settings, that there is less need to keep adding command line arguments for UI-things?

That's definitely something I've thought about. CLI options have grown to be something of a monster and persisted config is now my preferred way to deal with settings. 

I suppose the argument for CLI is advanced users who want to deploy a zero-config rollout of pikaraoke to their specification, say via docker. One could argue you could generate your own config.ini file to use in that case, but I'm not sure that's more or less convenient than CLI.

safepay:
Could I suggest that ones that are not crucial for command line start off as options in the UI and are then moved later to command line if user demands require it?

Otherwise removing a command line in the future is almost impossible.

vicwomg:
> Do you think its too late to remove the paging command arg?

Feel free to throw that in. But it's not urgent, if we want a general cleanup of non-crucial CLI args we can make a separate task in the future and have it bump the major version if we deprecate anything. But none of this feels urgent

safepay:
I removed the paging CLI. Not really related to the PR, but I think it's a good idea.

PR #679, opened 2026-01-18
feat: Update dependencies

safepay:
## Summary

**Phase 1 - Safe upgrades:**
- Remove unused CherryPy (migrated to Gevent)
- Babel 2.9.1 → 2.17.0
- requests 2.32.4 → 2.32.5 (security fix CVE-2024-47081)
- flask-paginate 2021.10.29 → 2024.4.12
- psutil 6.0.0 → 7.2.1
- qrcode 7.4.2 → 8.2 (using pypng instead of pillow for lighter dependencies)

**Phase 2 - Flask-Babel migration:**
- Flask-Babel 2.0.0 → 4.0.0
- Update locale selector from decorator to parameter pattern

## Note

Flask remains at 2.2.5 because flasgger is incompatible with Flask 3.x. Replacing flasgger with flask-smorest would unblock the Flask 3.x upgrade.

### DIFF
--- a/build_scripts/docker/Dockerfile
+++ b/build_scripts/docker/Dockerfile
@@ -30,4 +30,4 @@ HEALTHCHECK --interval=30s --timeout=3s --start-period=40s --retries=3 \
 
 USER pikaraoke
 
-ENTRYPOINT ["pikaraoke", "-d", "/app/pikaraoke-songs/", "--headless", "--config-file-path /app/config/config.ini"]
+ENTRYPOINT ["pikaraoke", "-d", "/app/pikaraoke-songs/", "--headless", "--config-file-path", "/app/config/config.ini"]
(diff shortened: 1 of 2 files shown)


safepay:
> It's great and much appreciated, but I'm really struggling to keep up with testing and am finding this is requiring much more of my time than I had anticipated. My goal of getting a release out keeps getting pushed back further and further.

I'll back off for a bit. I would like as a first pretty big step to both get the code base more maintainable (from your point-of-view) AND while getting to the goal of the SQLite db integration which will open up so many possibilities from a user's perspective.

The short term core-code goals are to update the dependencies, remove gevent and use native async, refactor the God class and then get to the SQLite update.

Then there are some fixes such as the YT previews not working if the video owner blocks them and some ideas around improving the YT search process, add backgrounds to MP3/CDG playback and maybe a couple of others that I would like to get done.

I'll add these as features so you can prioritise them and work out what the next release could look like.

But there is no rush.

PR #682, opened 2026-01-19
fix: prevent autoplay race condition and improve autoplay detection

safepay:
### Problem
After #670, the permissions modal would intermittently appear due to a race condition: `handleNowPlaying()` calls `playBGMusic()` before `testAutoplay()` completes, exhausting the browser's autoplay budget and causing the subsequent autoplay test to fail.

### Solution
1. **Race condition fix**: Added `autoplayConfirmed` flag - `playBGMusic()` and `playBGVideo()` return early until `handleConfirmation()` runs
2. **Improved autoplay test**: Start muted (always allowed), play, then unmute to check if browser permits audio - avoids the `NotAllowedError` that was corrupting browser state

### DIFF
--- a/pikaraoke/templates/splash.html
+++ b/pikaraoke/templates/splash.html
@@ -51,6 +51,7 @@
   var octopusInstance = null;
   var showMenu = false;
   var menuButtonVisible = false;
+  var autoplayConfirmed = false;
   var volume = 0.85;
   var playbackStartTimeout = 10000;
   var isScoreShown = false;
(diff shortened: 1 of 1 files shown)


vicwomg (approved):

vicwomg (approved):

vicwomg:
I merged this a little too soon. I noticed that this change seems to break first-time bg music play. It plays for a split second then shuts off.

vicwomg:
> I merged this a little too soon. I noticed that this change seems to break first-time bg music play. It plays for a split second then shuts off.

Specifically on safari. Looks like Chrome and Firefox are fine.

PR #684, opened 2026-01-19
Fix browser launch check and pause queue if no player present.

### LINKED ISSUE #681: Issue: Queue playback fails when no player client is connected
## Problem

When songs are added to the queue before the splash screen/player is active, playback fails with "Stream was not playable! Skipping track" after a 10-second timeout.

## Root Cause

The system doesn't verify that a player client is connected before attempting playback. It broadcasts `now_playing` via socket and hopes someone is listening.

### Flow:
1. User adds song to queue (via web UI on phone/computer)
2. Server starts processing the queue in `karaoke.run()` loop
3. Server transcodes the file and signals "Stream ready!"
4. Server sends `now_playing` via socket to splash page
5. Server waits 10 seconds for `is_playing` to become true (stream_manager.py:321-324)
6. **If no browser/splash page is listening** (not launched yet, crashed, or can't find screen), `start_song` socket event is never emitted by client
7. After 10 seconds timeout: "Stream was not playable! Skipping track"

## Additional Bug Found

In `app.py:250-252`, the browser launch failure check is ineffective:

```python
browser = Browser(k, args.window_size, args.external_monitor)
browser.launch_splash_screen()
if not browser:  # <-- Always truthy! browser is a Browser instance, not None
    logging.error("Failed to launch splash screen browser")
    sys.exit()
```

Should check `browser.browser_process` instead.

## Potential Solutions

1. **Track connected player clients** - Only attempt playback when at least one splash page client is connected via socket
2. **Queue songs but defer playback** - Don't start transcoding until a player is connected
3. **Fix browser launch check** - Check `browser.browser_process is None` after launch attempt
4. **Increase timeout or make it configurable** - For very slow hardware (though this only masks the issue)

## Related Files

- `pikaraoke/lib/stream_manager.py:320-330` - Wait loop for playback confirmation
- `pikaraoke/app.py:246-254` - Browser launch and (broken) error check
- `pikaraoke/karaoke.py:653-656` - `start_song()` sets `is_playing = True`
- `pikaraoke/templates/splash.html:360` - Client emits `start_song` socket event

## Related Issues

- #477

safepay:
Resolves #681 

This PR makes the queue wait for an available player before starting.

It let's you pre-queue, then start playing once the player is fired up.

If the player dies, the current song is cleaned up and the queue is paused at that point.

### DIFF
--- a/pikaraoke/app.py
+++ b/pikaraoke/app.py
@@ -88,11 +88,7 @@ app.register_blueprint(splash_bp)
 app.register_blueprint(controller_bp)
 app.register_blueprint(nowplaying_bp)
 
-babel.init_app(app)
-socketio.init_app(app)
 
-
-@babel.localeselector
 def get_locale() -> str | None:
     """Select the language to display based on user preference or Accept-Language header.
 
@@ -117,6 +113,10 @@ def get_locale() -> str | None:
     else:
         locale = request.accept_languages.best_match(LANGUAGES.keys())
     return locale
+
+
+babel.init_app(app, locale_selector=get_locale)
+socketio.init_app(app)
 
 
 # Handle all the socketio incoming events here.
--- a/pikaraoke/karaoke.py
+++ b/pikaraoke/karaoke.py
@@ -17,6 +17,7 @@ from typing import Any
 
 import qrcode
 from flask_babel import _
+from qrcode.image.pure import PyPNGImage
 
 from pikaraoke.lib.download_manager import DownloadManager
 from pikaraoke.lib.ffmpeg import (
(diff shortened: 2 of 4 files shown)


vicwomg:
seeing weird stuff with this one after some cursory testing. 

- Launch headless
- Add 3 random songs via web interface
- Open splash screen, 1st song starts playing, OK
- Close splash, song terminates, OK
- Re-open splash, next song begins, is frozen at 0:00
- Song errors out, skips to next. Next song also frozen at 0:00

safepay:
I think this is related to the other changes that have been added to the latest revert PR. Please confirm.

PR #687, opened 2026-01-19
Revert "fix: prevent autoplay race condition and improve autoplay detection

safepay:
This reverts commit 37e3c92.

The original behavior is correct. Modern browsers maintain a per-site "autoplay budget" based on user engagement (Media Engagement Index). Sites with low engagement show the permissions modal - this is expected browser behavior, not a bug.

The modal appeared during development due to repeatedly loading the player page without playing videos, which depletes the autoplay budget. In normal usage, playing videos builds engagement and the modal won't appear. Note: there is no way to manually reset this budget - it recovers naturally through normal site usage.

The reverted changes attempted to work around Chrome's autoplay detection but broke Safari video playback in the process.

### DIFF
--- a/pikaraoke/templates/splash.html
+++ b/pikaraoke/templates/splash.html
@@ -51,7 +51,6 @@
   var octopusInstance = null;
   var showMenu = false;
   var menuButtonVisible = false;
-  var autoplayConfirmed = false;
   var volume = 0.85;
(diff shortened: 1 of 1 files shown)


vicwomg:
See my fix here. The autoplay detection function wasn't working properly: https://github.com/vicwomg/pikaraoke/pull/688

PR #688, opened 2026-01-19
fix: autoplay detection test positive, mobile layout issue

vicwomg:
- Autoplay detection was returning a false positive due to only checking muted, and not checking for paused state. Browsers will auto-pause an autoplay video when the permissions are not enabled
- Give browser some extra time to play the video and determine state
- Add a longer video to prevent it from completing before test is done
- Fix issue with mobile layout when showing iframe


pikaraoke/templates/splash.html
```diff
@@ -427,16 +427,11 @@
 
     // Test if autoplay with audio is allowed using a real video file
     const testAutoplay = async () => {
-      if (isForcePermissionsModalBrowser) {
-        // Uses hardcoded unsupported browsers. Use this for browsers that somehow pass the below test but still fail to autoplay
-        enablePermissionsModal();
-        return;
-      }
```
vicwomg: No longer need this because the test now works properly

safepay: Great. This has been giving me a hernia.

PR #693, opened 2026-01-20
fix: improve logging: yt-dlp upgrades and pikaraoke version

vicwomg:
- Move more logging to debug
- Don't try to upgrade yt-dlp if its current
- Log pikaraoke version on start


safepay:
As a suggestion, you can add a check for new version that appears in the info page when the is_admin flag is set. This would help users know when to upgrade.

PR #715, opened 2026-01-28
feat: splash screen url scaling

safepay:
## Summary
- Add `?scale=` URL parameter to `/splash` for scaling down UI elements
- Resolves #699 

## Usage

- Use `/splash` as normal (no parameter needed)
- Or, bookmark `/splash?scale=0.6` (adjust value between 0.3 and 1.0 as needed)

Scales overlay elements (logo, now playing, up next, QR code, etc.) while keeping video playback, background video, and menu at full size.

## Test plan
- [ ] Open `/splash` with no parameter - verify no visual change
- [ ] Open `/splash?scale=0.7` - verify UI elements are scaled down
- [ ] Open `/splash?scale=0.5` - verify smaller scaling
- [ ] Verify video playback is unaffected at any scale
- [ ] Verify screensaver DVD logo scales correctly
- [ ] Verify menu button and menu iframe remain full size

### DIFF
--- a/pikaraoke/static/js/splash.js
+++ b/pikaraoke/static/js/splash.js
@@ -18,6 +18,7 @@ var screensaverTimeoutSeconds = PikaraokeConfig.screensaverTimeout;
 var bg_playlist = []
 const scoreReviews = PikaraokeConfig.scorePhrases;
 var isMaster = false;
+var uiScale = null;
 
 // Browser detection
 const isSafari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent);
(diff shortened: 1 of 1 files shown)


vicwomg:
Not a huge deal but slight bug: the screensaver scales in size, but its bounds are limited to the scale box, so it doesn't extend the whole screen when bouncing around corners

safepay:
The fix for the score is to replace:
```
    { selector: '#score', origin: null },
```

with
```
    { selector: '#your-score-text', origin: null },
    { selector: '#score-number-text', origin: null },
    { selector: '#score-review-text', origin: null },
```

PR #717, opened 2026-01-28
fix: make yt-dlp updating a background task

safepay:
## Proof-of-concept: Fix yt-dlp upgrade reliability and startup performance

Feel free to comment/delete this PR.

### Changes

**youtube_dl.py:**
- Use `sys.executable -m pip` instead of hardcoded `pip3`/`pip` commands to ensure upgrades target the correct Python environment
- Make `--break-system-packages` conditional - only add when NOT in a virtual environment (checks `sys.prefix == sys.base_prefix`)
- Improved error logging with exception details

**app.py:**
- Make yt-dlp upgrade async using `gevent.spawn()` to prevent blocking application startup
- Clean up redundant gevent import

### Why

The previous approach had two issues:

1. **Reliability**: Hardcoded `pip3`/`pip` commands could upgrade the wrong Python environment, especially in virtualenvs or pipx installations
2. **Performance**: Synchronous upgrade call blocked app startup, delaying server availability by several seconds

These changes ensure upgrades work correctly across all installation methods (system, venv, pipx) and allow the app to start immediately while upgrades happen in the background.

### DIFF
--- a/pikaraoke/lib/download_manager.py
+++ b/pikaraoke/lib/download_manager.py
@@ -211,7 +211,6 @@ class DownloadManager:
         k.log_and_send(_("Downloading video: %s") % displayed_title)
 
         cmd = build_ytdl_download_command(
-            k.youtubedl_path,
             video_url,
             k.download_path,
             k.high_quality,


pikaraoke/lib/youtube_dl.py
```diff
@@ -131,19 +131,20 @@ def upgrade_youtubedl(youtubedl_path: str) -> str:
                 pass
 
         if not upgrade_success:
-            # allow pip to break system packages (probably required if installed without venv)
-            args = ["install", "--upgrade", "yt-dlp[default]", "--break-system-packages"]
+            pip_cmd = [sys.executable, "-m", "pip", "install", "--upgrade", "yt-dlp[default]"]
+
+            # Outside a venv, pip requires --break-system-packages on modern Python
+            if sys.prefix == sys.base_prefix:
+                pip_cmd.append("--break-system-packages")
```
vicwomg: this is great, I was actually trying to figure out how to make this work but I gave up, instead I made the install scripts install yt-dlp globally. That means you can probably remove any code in the installers that does `pipx install yt-dlp`

I believe you can also remove lines 119-131 from this file now, because presumably the sys.executable will handle it more correctly.

Issue #672, opened 2026-01-16
[REFACTOR] Karaoke Class Refactoring Plan


safepay:
# Karaoke Refactoring Master Plan

## Strategic Motivation

- **Single-Maintainer Efficiency:** Decoupling isolates complexity, allowing you to work on specific components (e.g., Queue) without needing to hold the entire system state in memory.
- **Test Simplicity:** Shifts from complex mocking to using real, lightweight instances (Events, Preferences) in tests, reducing test maintenance.
- **Structural Clarity:** Eliminates circular dependencies and complex control flow by enforcing a clean Coordinator pattern.

## 1. Architectural Objective

Transition `Karaoke.py` from a central class containing all logic into a **Coordinator** that wires together independent, testable **Managers**.

**Communication Strategy:**

- **Direct calls** where possible (Manager -> Coordinator is fine)
- **Simple callbacks** for async operations (downloads, playback end)
- **Minimal EventSystem** only where SocketIO updates are needed (avoids passing SocketIO everywhere)

The EventSystem is intentionally kept to ~15 lines - just a callback registry, nothing fancy. It exists solely to avoid passing `socketio` and `flash` objects deep into managers.

**Error Handling:**

- Managers raise exceptions for errors
- Coordinator catches and handles them (logging, UI notifications)
- EventSystem handlers are called synchronously - exceptions bubble up normally

**Migration Strategy:**

- Each PR is backwards compatible
- Existing config.ini files are migrated on first load
- No data loss, no user action required

______________________________________________________________________

## 2. Consolidated Project Roadmap

| Stage / PR | Component | Description | Status |
| :--- | :--- | :--- | :--- |
| **PR #711** | Infrastructure | Initial setup, socketio tests | **Merged** |
| **PR #712** | QueueManager | Initial extraction | **Merged** |
| **PR #727** | **Foundation** | **Create PreferenceManager + EventSystem** (Critical first step) | **Merged** |
| **PR [#758](https://github.com/mannr/pikaraoke/pull/758)** | QueueManager | **Cleanup:** Inject Prefs/Events, **Rewrite Tests** to use Real Events | **Merged** |
| **PR [#768](https://github.com/mannr/pikaraoke/pull/768)** | Playback | **PlaybackController:** Extract logic, use Event-driven architecture | **Merged** |
| **PR [#771](https://github.com/mannr/pikaraoke/pull/771)** | Library | **SongManager:** Centralize song/file operations & search | **Merged** |
[...]

PR #727, opened 2026-01-29
refactor: move preference handling to a new file

safepay:
## PR3: Foundation - PreferenceManager + EventSystem from #672

**Goal:** Establish single source of truth for configuration and minimal communication backbone for future decoupling.

### What Changed

**New Infrastructure Components:**
- **`pikaraoke/lib/events.py`** - Minimal 25-line event dispatcher providing clean callback registry
  - Simple `on(event, handler)` and `emit(event, *args)` API
  - Synchronous execution - no async complexity
  - Will replace direct socketio/flash callbacks in PR4+
  - **Status:** Created but not yet wired up (no `.on()` or `.emit()` calls yet)
  
- **`pikaraoke/lib/preference_manager.py`** - Centralized configuration management (~146 lines)
  - Single `DEFAULTS` dict - one source of truth for all settings
  - Auto type conversion (bool/int/float) from config.ini strings
  - Legacy config.ini migration from cwd to data directory
  - **Status:** Fully integrated and actively used

**Core Integration:**
- **`pikaraoke/karaoke.py`** - Major cleanup and simplification
  - Initialize `self.events = EventSystem()` and `self.preferences = PreferenceManager()`
  - Replaced all direct `config.ini` reads with `self.preferences.get()`
  - New `_load_preferences()` method consolidates config loading logic
  - Reduced complexity in `__init__()` - clearer initialization flow
  
**Route Updates:**
- `pikaraoke/routes/preferences.py` - Uses PreferenceManager for get/set operations
- `pikaraoke/routes/info.py` - Uses PreferenceManager for system info
- `pikaraoke/routes/files.py` - Uses PreferenceManager for file operations
- `pikaraoke/app.py` - Updated initialization

**Test Coverage:**
- `tests/unit/test_events.py` - Event registration and emission behavior
- `tests/unit/test_preference_manager.py` - Config persistence, type conversion, migration
- `tests/conftest.py` - Fixtures updated for new structure
- `tests/unit/test_karaoke_utils.py` - Updated for PreferenceManager integration


safepay:
At the end of the day it's your policy decision:

Option A - set the config.ini from the CLI as am "initial default". If it gets changed in the UI, then passing the CLI in a future restart does nothing. This could also log via "info" that the CLI was not applied due to a config setting and to adjust it there.

Option B - every time the system is restarted the CLI value overwrites the config.ini setting.

Honestly, I can see pros and cons either way. I would probably do Option A, because if I changed server behaviour to my liking via the UI, I would expect that on a restart. This is why the CLI arguments are somewhat redundant these days for UI settings.

vicwomg:
I think option B is best. I can't think of a way to make Option A work consistently and seems just as complicated as before. How do you tell when an option has indeed "been changed in the UI"?

If you are savvy enough to use CLI args, then chances are you know they will clobber your UI settings on restart. Most users will never us them. There can even be some messaging in the --help copy to explain this.

vicwomg:
think of it this way: CLI means "these are my initial settings on launch". What changes after will persist for the session, but if I use those CLI args again, the initial settings are restored.

safepay:
Ready for your review.

PR #771, opened 2026-02-12
refactor: song manager

safepay:
Addresses the next step in #672

## What Changed

**Song library operations have one owner instead of living in the Karaoke god class.**

- Extracted `SongManager` -- single place for all song discovery, delete, rename, and display name logic. When a file operation misbehaves, you open one 79-line file instead of scrolling through Karaoke's 400+ lines.
- `SongManager` owns the `SongList` instance (`songs`) and the `download_path`. Routes now access `k.song_manager.songs` instead of `k.available_songs`, and `k.song_manager.delete()` instead of `k.delete()`.
- `filename_from_path` is now a `@staticmethod` on `SongManager`. It never used instance state -- making it static makes that explicit and lets callers (Jinja templates, QueueManager callbacks) use it without a Karaoke reference.
- `rename()` simplified: removed the `len(ext) == 2` guard that could never be false (splitext always returns a 2-tuple).

**Cleanup**

- Fixed `SongList.scan_directory()` and `find_by_id()` to use `str(path)` instead of `path.as_posix()`. On Windows, `as_posix()` produces forward-slash paths that don't match `os.path` operations elsewhere, causing songs to be unfindable.
- Tests use real filesystem (`tmp_path`) with no mocks -- delete, rename, and companion file handling are verified against actual files.
- `TestFilenameFromPath` moved from `test_karaoke_utils.py` to `test_song_manager.py`, testing the static method directly instead of through a mock Karaoke instance.

Clear boundaries: SongManager owns song files, QueueManager owns queue, PlaybackController owns playback state, Karaoke coordinates.

---

## Manual Test Plan

**Files modified:** `song_manager.py` (new), `song_list.py`, `karaoke.py`, `download_manager.py`, 5 route files, all relevant tests

Focus: verify song library operations work through the new ownership path. Playback and queue logic are unchanged.

### Song Library (scan/refresh moved to SongManager)
- [x] Songs load on startup -- browse page shows full library
- [x] Admin refresh re-scans directory and picks up new files
- [x] Autocomplete search returns matching songs

### File Operations (delete/rename moved to SongManager)
- [x] Delete song from browse page -- file removed, song disappears from library
- [x] Delete song with .cdg companion -- both files removed
- [x] Delete song with .ass subtitle -- both files removed
- [x] Rename song from edit page -- file renamed, library updated
- [x] Rename song with companion files -- companions renamed too
- [x] Rename to existing filename shows error (not silent overwrite)

### Display Names (filename_from_path now static on SongManager)
- [x] Browse page shows clean names (no extensions, no YouTube IDs)
- [x] Queue page shows clean song names
- [x] Search autocomplete shows clean names
- [x] Batch renamer shows correct current vs suggested names

### Batch Song Renamer (routes updated to use SongManager)
- [x] Batch renamer loads all songs with pagination
- [x] Batch renamer filters to songs needing rename
- [x] Batch rename executes correctly

### Download Integration (DownloadManager references updated)
- [x] Download a song -- appears in library after completion
- [x] Download with auto-enqueue -- song added to queue after download

---

Unit tests: `test_song_manager.py` (new), `test_download_manager.py` (updated), `test_karaoke_utils.py` (trimmed). All pass via `uv run pytest`.

No user-facing behavior changes. All song operations should work identically to before.

### DIFF
--- a/pikaraoke/lib/song_manager.py
+++ b/pikaraoke/lib/song_manager.py
@@ -5,8 +5,20 @@ from __future__ import annotations
 import contextlib
 import logging
 import os
+import re
 
+from pikaraoke.lib.get_platform import is_windows
 from pikaraoke.lib.song_list import SongList
+
+# Characters illegal in Windows filenames
+_WINDOWS_ILLEGAL_CHARS = re.compile(r'[<>:"/\\|?*]')
+
+
+def sanitize_filename(name: str) -> str:
+    """Remove characters that are illegal in filenames on the current platform."""
+    if is_windows():
+        name = _WINDOWS_ILLEGAL_CHARS.sub("-", name)
+    return name.strip()
 
 
 class SongManager:
@@ -68,6 +80,7 @@ class SongManager:
             song_path: Full path to the current song file.
             new_name: New filename (without extension).
         """
+        new_name = sanitize_filename(new_name)
         logging.info(f"Renaming song: '{song_path}' to: {new_name}")
         companions = self._get_companion_files(song_path)
         _, ext = os.path.splitext(song_path)
--- a/pikaraoke/routes/files.py
+++ b/pikaraoke/routes/files.py
@@ -1,5 +1,6 @@
 """File management routes for browsing, editing, and deleting songs."""
 
+import logging
 import os
 import unicodedata
 from urllib.parse import unquote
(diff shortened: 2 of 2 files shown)


vicwomg:
You can assign to me for review when you think its ready

PR #772, opened 2026-02-12
fix: browser and edit page navigation improvements

safepay:
Two small changes to simplify navigation and song editing

* Added icon on browse page to allow all songs to be listed after sublisting via letters
* Added navigation from the edit page back to the referrer page where edit was clicked


safepay:
Ha. I missed one. All done now!

vicwomg:
> With a lot of the navigation now working more seamlessly, I would consider dropping the default results per page to 100 instead of 500 to make it more mobile friendly. Even 50 would be ok.

I think 100 is ok, I believe that's what it was before all the refactoring

safepay:
> I think 100 is ok, I believe that's what it was before all the refactoring

It was always 500 which I found pretty high. I never changed that. But I'll set it at 100 now.

Issue #691, opened 2026-01-20
[FEATURE] Update Dependencies Phase 3 - flasgger replacement


safepay:
# Dependency Upgrade Evaluation for PiKaraoke

## Summary Table

| Dependency      | Current    | Latest  | Action              | Notes                          |
|-----------------|------------|---------|---------------------|--------------------------------|
| CherryPy        | 18.10.0    | -       | REMOVED          | Unused (migrated to Gevent)    |
| Flask           | 2.2.5      | 3.1.2   | **UPGRADE**    | **Blocked by Flasgger**           |
| Jinja2          | 3.1.6      | 3.1.6   | Current             | -                              |
| Babel           | 2.9.1      | 2.17.0  | UPGRADED         | CLDR 46, Python 3.12+          |
| Flask-Babel     | 2.0.0      | 4.0.0   | UPGRADED         | Requires code change           |
| requests        | 2.32.4     | 2.32.5  | UPGRADED         | Security fix (CVE-2024-47081)  |
| flask-paginate  | 2021.10.29 | 2024.4.12 | UPGRADED       | Bug fixes                      |
| ffmpeg-python   | 0.2.0      | 0.2.0   | Keep         | Unmaintained since 2019        |
| flask-socketio  | >=5.5.0    | 5.6.0   | Current             | -                              |
| gevent          | >=24.11.1  | 25.9.1  | Current             | Replace with asyncio direct calls                              |
| flasgger        | >=0.9.5    | 0.9.5   | **REPLACE**  | **Blocks Flask 3.x**               |
| psutil          | 6.0.0      | 7.2.1   | UPGRADED         | Performance, Python 3.14       |
| qrcode          | 7.4.2      | 8.2     | UPGRADED         | Bug fixes                      |

---

### flasgger - Blocks Flask 3.x

**Problem:** Incompatible with Flask 3.x (uses removed `JSONEncoder`). Unmaintained.

**Recommended replacement:** flask-smorest

- Actively maintained (v0.46.2, Sep 2025), Flask 3.x compatible
- Requires Marshmallow schemas instead of YAML docstrings
- Medium effort: define schemas for ~14 route files, update decorators
- Replace --enable-swagger with --enable-apidocs (or use both for compatibility)

---

## Code Changes Required
- Replace YAML with Marshmallow schemas and decorators.
- Update the routes to use the new schemas
- Ensure /apidocs render the same

---

### Phase 1: Safe Upgrades (No Code Changes) COMPLETED

[...]

PR #775, opened 2026-02-17
refactor: replace flassger with flask smorest

safepay:
This addresses #691.

This is for discussion noting that:
-  flask-smorest exposes ALL routes and cannot hide specific routes - therefore there are more api endpoints than before
-  if we want to hide non-json endpoints, this requires splitting the routes in to two files: one for html and one for api (see https://github.com/safepay/pikaraoke/tree/refactor-split-html-api-routes for comparison)
-  there were several alternatives explored but they were not clean and simple
- The old apidocs _already_ had html endpoints, so adding some more does not significantly change the behaviour.
- /apidocs is an edge case: it's only used by one or maybe a few more users and only when explicitly enabled. It is not core behaviour, just a mainenance headache.

- This branch is the _simplest way_ to remove flasgger and create a managable code base that automatically documents the api end points.

**Recommendation:**
- **Merge this branch and accept the caveats.**

### DIFF
--- a/pikaraoke/routes/batch_song_renamer.py
+++ b/pikaraoke/routes/batch_song_renamer.py
@@ -380,7 +380,7 @@ def get_song_correct_name(song):
     return get_best_result(results, cleaned_query)
 
 
-@batch_song_renamer_bp.route("/batch-song-renamer", methods=["GET", "POST"])
+@batch_song_renamer_bp.route("/batch-song-renamer", methods=["GET"])
 def browse():
     """Batch song renamer page."""
     if not is_admin():


safepay:
Not yet ready for review

safepay:
### Replace flasgger with flask-smorest

**Why replace flasgger?**

flasgger generates OpenAPI 2.0 (Swagger) docs from YAML blocks embedded in docstrings. This approach has problems:
- YAML docs drift out of sync with the actual `request.args`/`request.form` calls
- No validation — parameters are documented but not enforced
- Verbose — each route needed 10-15 lines of manually maintained YAML
- flasgger is pinned to OpenAPI 2.0, which is end-of-life

**What flask-smorest gives us:**

- **OpenAPI 3.0.2** — the current standard
- **Marshmallow schemas for request parameters** — defines the parameter name, type, default, and description in one place. The same schema that documents the API also parses and validates incoming data, so docs can't drift from reality
- **Less boilerplate** — a 4-line schema class replaces 10-15 lines of YAML
- **Validation for free** — type coercion, defaults, and required checks happen automatically instead of manual `request.args.get("page", type=int, default=1)` calls scattered through route functions
- **Cleaner Flask integration** — flask-smorest extends Flask's `Blueprint`, so the migration is mostly mechanical: swap the import, add schemas, remove YAML blocks

**Blueprint registration:**

Blueprints are now split into two groups in `app.py`:
- `_api_blueprints` — registered via `api.register_blueprint()`, visible in `/apidocs`
- `_internal_blueprints` — registered directly with Flask, hidden from docs (UI-only routes like home, splash, stream)

This replaces flasgger's per-route `tags:` YAML with a structural separation that's harder to misconfigure.

**Form field naming:**

HTML form field names changed from kebab-case (`song-to-add`) to snake_case (`song_to_add`) to match marshmallow field naming conventions. Affected templates updated accordingly.

**Dependency changes:**

- Removed: `flasgger`, pinned `Flask==2.2.5`, pinned `jinja2`
- Added: `flask-smorest>=0.46.0`, `Flask>=3.1.0,<4` (flask-smorest requires Flask 3+)

safepay:
Note that the overarching issue is the need to keep supporting apidocs while the code base mixes html and json endpoints.
This creates a difficult situation.
 
These changes, while seemingly significant, stop short of splitting the html and json routes into separate files.

The challenge is that with apidocs there comes a certain level of documentation which is unavoidable.

I have tried to find the best balance between minimising the PR, making the code base maintainable, while having to keep apidocs working.

PR #802, opened 2026-03-12
fix: add admin security around edit, delete and addrandom routes

safepay:
## Problem
No authentication was wrapped around file or queue admin management functions.

## Solution
Add `is_admin` around `queue/addrandom`, `/files/edit` and `/files/delete` routes

This is good practice and could help with resolving #801


safepay:
Delay this until the other two PR's are merged, as this one is less important.

Issue #641, opened 2026-01-06
[FEATURE] SQLite Metadata Layer & Background Enrichment


safepay:
### Current Situation & Problem

* **Messy Data:** Search relies entirely on raw filenames, which are often inconsistent (e.g., `vocal_star_waterloo_guide.mp3`).
* **Missing Metadata:** Impossible to filter by **Decade** or **Genre** as this data does not exist in filenames.
* **Duplicate Entries:** Song variants (e.g., "Vocal Guide", "Acoustic") appear as duplicate entries rather than grouped versions.
* **Feature Blocking:** The lack of persistent song IDs makes it impossible to implement future features like **Playlists**, **Favorites**, or **History**.

### Proposed Solution

Move to a `Filesystem -> SQLite -> RAM` architecture to decouple file storage from user display.

* **SQLite Database:** Acts as the persistent "System of Record" for file paths, cleaned metadata, and variant tags.
* **RAM Index:** The app loads clean data from the DB into memory on startup, preserving instant search performance.
* **Platform Performance:** Significantly reduces startup time on hardware with limited I/O (e.g., Raspberry Pi 3B) by eliminating the need to crawl thousands of files on every boot.
* **Non-Blocking Updates:** The Admin "Rescan" button updates the DB instead of blocking the playback engine.
* **Enrichment:** A separate background process queries external APIs to clean titles and fetch metadata (Year, Genre).

### Implementation Plan

* **Database Infrastructure:** Create a `pikaraoke_db` module with a schema supporting `artist`, `title`, `variant`, `year`, and `genre`.
* **Smart Scanner:** Update scanning logic to automatically pair files (MP3+CDG, MP4+ASS), support ZIPs, and extract variant tags (e.g., `[Vocal Guide]`) from filenames.
* **App Integration:** Modify the main application to load the song list from the database instead of directly crawling the filesystem.
* **Background Enrichment:** Create scripts to query Last.fm (for cleaning & Genre tags) and MusicBrainz (for Year) to populate the new database columns.

poeticize:
I agree, which I why the existing feature add I did has no backend change requirements and merges without issue. User data elements from this update add a nice management feature and can easily be removed from the single search page.  I will definitely help maintain this feature and propose a clean move of its functions into the DB once it is present.

safepay:
Closed as completed. The metadata lookup will be handled separately.

PR #804, opened 2026-03-19
feat: replace lastfm with itunes on song edit page and improve UI

safepay:
## Context

Last.fm has poor metadata quality (wrong capitalisation, artist name embedded in title) and relies on a hardcoded API key belonging to an individual user, shared silently by every PiKaraoke install. This PR replaces it on the edit page with the iTunes Search API (#641), which is free and requires no API key, and returns canonical artist names, correct capitalisation, release year, and primary genre from a curated commercial catalogue.

This is a **stepping stone toward background metadata enrichment**. The database schema (merged separately in #792) already includes `metadata_status`, `enrichment_attempts`, and `last_enrichment_attempt` columns. A future PR will wire up a background worker that automatically enriches songs with correct title, artist, year, and genre — and once that browse-based workflow proves itself, the batch song renamer (which still uses Last.fm) can be retired in favour of browse-page status filtering (e.g. `/browse?status=failed`).

## Changes

- **iTunes metadata provider** (`metadata_providers.py`): `MetadataProvider` protocol, `ITunesProvider` with rate limiting and retry logic, `suggest_metadata()` pipeline, `get_provider()` factory — designed for future providers (e.g. Spotify)
- **Improved metadata parsing**: accent-insensitive matching, featuring-artist preservation in brackets, expanded special version keywords (remastered, deluxe, demo, etc.)
- **Edit page UX refresh**: suggestions now show year, genre, and match score; removed separate "tidy name" button in favour of integrated flow
- **Format icons**: file type badges (mp4, cdg, mkv, webm, etc.) for the browse page for future PR
- **`metadata_provider` preference key** added to `PreferenceManager`

## Testing

To validate the iTunes provider ahead of background enrichment (which will process the full library automatically), run the bulk report script against your song directory:

```bash
uv run python scripts/itunes_bulk_report.py /path/to/your/songs
```

This produces `scripts/itunes_report.csv` with columns: original filename, tidied name, match score, suggested display, artist, title, year, and genre — plus a console summary showing score distribution. Processing is ~3s per song due to iTunes rate limits, which is fine for background enrichment but means a large library will take a while to report on.

### Manual test plan

- [x] Open the edit page, click "Search" and confirm results show title, artist, year, genre, and score from iTunes
- [x] Click a suggestion and verify the rename input is populated correctly
- [x] On the info page, verify song count displays and "Sync Now" triggers a re-scan with SocketIO feedback


vicwomg:
When I hit "search" in the edit screen, nothing happens and I see this in the logs:

```

[2026-03-21 23:42:54] ERROR: Exception on /metadata/suggest-names [GET]
Traceback (most recent call last):
  File "/Users/vic/coding/pikaraoke/.venv/lib/python3.11/site-packages/flask/app.py", line 1511, in wsgi_app
    response = self.full_dispatch_request()
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vic/coding/pikaraoke/.venv/lib/python3.11/site-packages/flask/app.py", line 919, in full_dispatch_request
    rv = self.handle_user_exception(e)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vic/coding/pikaraoke/.venv/lib/python3.11/site-packages/flask/app.py", line 917, in full_dispatch_request
    rv = self.dispatch_request()
         ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vic/coding/pikaraoke/.venv/lib/python3.11/site-packages/flask/app.py", line 902, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vic/coding/pikaraoke/.venv/lib/python3.11/site-packages/webargs/core.py", line 652, in wrapper
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vic/coding/pikaraoke/.venv/lib/python3.11/site-packages/flask_smorest/arguments.py", line 85, in wrapper
    return current_app.ensure_sync(func)(*f_args, **f_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vic/coding/pikaraoke/pikaraoke/routes/metadata_api.py", line 24, in suggest_names
    results = suggest_metadata(query["filename"], provider=provider, limit=query["limit"])
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vic/coding/pikaraoke/pikaraoke/lib/metadata_providers.py", line 450, in suggest_metadata
    results = provider.search(search_query, limit=limit * _OVERFETCH_FACTOR)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vic/coding/pikaraoke/pikaraoke/lib/metadata_providers.py", line 138, in search
    response = requests.get(
               ^^^^^^^^^^^^^
  File "/Users/vic/coding/pikaraoke/.venv/lib/python3.11/site-packages/requests/api.py", line 73, in get
    return request("get", url, params=params, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vic/coding/pikaraoke/.venv/lib/python3.11/site-packages/requests/api.py", line 59, in request
    return session.request(method=method, url=url, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vic/coding/pikaraoke/.venv/lib/python3.11/site-packages/requests/sessions.py", line 589, in request
    resp = self.send(prep, **send_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vic/coding/pikaraoke/.venv/lib/python3.11/site-packages/requests/sessions.py", line 703, in send
    r = adapter.send(request, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vic/coding/pikaraoke/.venv/lib/python3.11/site-packages/requests/adapters.py", line 644, in send
    resp = conn.urlopen(
           ^^^^^^^^^^^^^
  File "/Users/vic/coding/pikaraoke/.venv/lib/python3.11/site-packages/urllib3/connectionpool.py", line 787, in urlopen
    response = self._make_request(
               ^^^^^^^^^^^^^^^^^^^
  File "/Users/vic/coding/pikaraoke/.venv/lib/python3.11/site-packages/urllib3/connectionpool.py", line 464, in _make_request
    self._validate_conn(conn)
  File "/Users/vic/coding/pikaraoke/.venv/lib/python3.11/site-packages/urllib3/connectionpool.py", line 1093, in _validate_conn
    conn.connect()
  File "/Users/vic/coding/pikaraoke/.venv/lib/python3.11/site-packages/urllib3/connection.py", line 796, in connect
    sock_and_verified = _ssl_wrap_socket_and_match_hostname(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vic/coding/pikaraoke/.venv/lib/python3.11/site-packages/urllib3/connection.py", line 928, in _ssl_wrap_socket_and_match_hostname
    context = create_urllib3_context(
              ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/vic/coding/pikaraoke/.venv/lib/python3.11/site-packages/urllib3/util/ssl_.py", line 303, in create_urllib3_context
    context.minimum_version = TLSVersion.TLSv1_2
    ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/ssl.py", line 608, in minimum_version
    super(SSLContext, SSLContext).minimum_version.__set__(self, value)
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/ssl.py", line 608, in minimum_version
    super(SSLContext, SSLContext).minimum_version.__set__(self, value)
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/ssl.py", line 608, in minimum_version
    super(SSLContext, SSLContext).minimum_version.__set__(self, value)
  [Previous line repeated 474 more times]
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/ssl.py", line 606, in minimum_version
    if value == TLSVersion.SSLv3:
                ^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/enum.py", line 196, in __get__
    return ownerclass._member_map_[self.name]
           ^^^^^^^^^^^^^^^^^^^^^^^
RecursionError: maximum recursion depth exceeded while calling a Python object
```

safepay:
I have been playing with the UI, so "Auto Suggest" became "Search".

The exception is a bug in Python 3.11. There is already a fix in place for this recursion bug for Python 3.13. I'll amend the code to handle earlier versions as well.

vicwomg:
Ok, ran on my library. I think the main trip up is that any prefix string that is not included in the title is going to throw it off. Would probably suffice to just to strip any "KARAOKE" string.

Also I'd like to see the Auto-format feature return for search misses. Perhaps a button next to swap artist title.

vicwomg:
Playing with this more, I am quite lost. The new auto-tidier feature changes displayed filenames superficially so it's not clear what I am renaming or if it has been done already. I'm really losing the plot.

safepay:
The regex_tidy functions are now merged and are optional via the checkbox on info.html
If there are any changes required to the parsing, these should be done in a **separate PR**.

This branch is now solely concerned with the iTunes work.
Noth that it STILL tidies the file name in the search box, regardless of the checkbox in info.html, because there is no auto-format.

If you want auto-format, then we can also disable automatic tidying on that page (based on the setting) and let the user "auto-format" by applying the tidy function themselves.

But you are making me think that this whole approach is a bridge too far for PiKaraoke and if something more simplistic should be done instead.

vicwomg:
I think the auto-tidying of the edit field is throwing me. Any sort of opaque automatic action will ultimately lead to confusion. It should probably be a manual step so I know what is happening. Also the edit field paired with the search box is weird because it seems to imply you're editing a search field to send to the search (inline search button), not the filename itself to save (below). I believe the previous UI was clearer. In this case it would be "suggest from iTunes", which presents a list that would edit the box. Auto-format, swap song/artist, suggest from iTunes are all operations that modify the file name, which then must be saved. 

Ultimately the goal is to have artist - song display come from one source, the db, right? I think the more we try to hack around file renaming it's going to get complex. We need less ambiguity as to what song title is displayed. If it''s decoupled from filename, all the better, but supporting both and auto-tidying superficially is getting extremely complex and unclear to the end user.

safepay:
The issue is that dealing with previous PR's that introduced different, but overlapping code - edit.html and batch_renamer, have created a huge head-ache on how to handle the same process in a centrally managed way. Those PR introduced some challenging concepts such as personal API keys within the code and name handling that was specific to one person's song library. Plus the problem of handling dirty data from LastFM.

This has been very hard to unravel. This work has been trying to do that, but in itself has become entangled.

From this point on we can look at is as two clear but separate issues.

1. Tidying names that come from YouTube via code - already merged but can be cleaned up more
2. Looking up names via an API to get clean information - this PR

vicwomg:
Sounds to me that file renaming in general needs to die. edit.html + the batch editor edits the artist - song fields in the db. The song list pulls db data for display (falls back to filename if those are null) while the backend always uses the filename for identification.

safepay:
> Sounds to me that file renaming in general needs to die. edit.html + the batch editor edits the artist - song fields in the db. The song list pulls db data for display (falls back to filename if those are null) while the backend always uses the filename for identification.

This is certainly adding to the confusion. The "edit" page is ideally there to manage db metadata for a song. It could, in the future optionally also rename the file from the metadata.

Because this code is a bridge between these cases, it makes the transtion difficult to manage, hence you losing the plot. I f we factor temporary insanity in to the mix, it makes it quite hard, don't you think?

safepay:
Seriously though, the issue is when to switch to database metadata vs filenames.

That is the idea of the background worker. It gets most of the job done behind the scenes. Then the user can tidy up by browsing a select list of songs that have failed the automated process and using the edit screen. This reuses the browse page and therefore eliminates the batch_renamer.

That is why the regex_tidy and API process is so crucial.

vicwomg:
I don't think automated processes for something as unreliable as song identification is a good idea. Nothing is accounting for wrong matches that were written because the system deemed them correct. If there is to be any kind of mass-metadata editing (or any editing to sensitive data at all) it should be presented to the user for review before writing. 

Many users are perfectly happy with unclean titles. At least they contain all the information as they downloaded it and can find it. I don't think the system should be opinionated in this way.

safepay:
I'll do some work on the UI when I get a chance and also do a series of smaller PR's that will allow the display of songs to use the metadata fields for title and artist as well as the filename as a fallback. That way, we can start using the power of the db while transitioning from filenames to metadata. Let me know your thoughts.

safepay:
> Many users are perfectly happy with unclean titles. At least they contain all the information as they downloaded it and can find it. I don't think the system should be opinionated in this way.

Maybe there needs to be a switch to "Use Title and Artist metadata" or something if people are happy with the current system.
Otherwise there is not much point putting data in those fields.

And then you could optionally run the background process as well. If you don't like the results, just turn it off and go back to the filenames.

vicwomg:
Sounds good, I think if we can get file renaming with iTunes suggestion UIs
a little more clear then this PR can go in. Definitely valuable in the
short term and removes dependency on tokens and lastfm.

The main thing is it’s not clear if this is an edit box or a search box and
what in fact that search does. I think decoupling the text fields from the
actions that act on it is the way to go. Not sure how to best handle the
country code thing, but I’ll let you know if I think of something.

safepay:
This should be getting closer now.

Regarding the bulk updating of songs, I have a solution that would involve activating a background worker by an admin button click which would then store the suggestions in the db in "suggestion" columns. A user would then be able to use a modified batch_renamer to review the suggestions and accept/edit as appropriate.

The metadata could then live with the filename at the same time but take precedence if it exists.

vicwomg:
The UI looks good to me, if there's nothing left to do I can merge

safepay:
Ok. Good to go.

safepay:
And I've done some thinking on both modifying the edit page to populate the artist/title/year fields as well as the bulk updating process which is worthy of some discussion and can be moved to the Issues register. But first to get your thoughts.

1. Starting to use the metadata.

- This is not too hard as it requires modifying the edit page (just UI)
- Then modifying the search and browse to look at both the filename and the title/artist fields at the same time.
- The complication is that the system does not know which way around a user prefers their filenames - artist or title first
- So we have to add a config flag that asks and sets this.
- Then the browse will order alphabetically by the filename and the artist OR title, if that metadata is available, and flip the order of title and artist metadata accordingly.

2. The bulk metadata update process

- The concern is managing a slow api as well as a user's ability to confirm each change.
- The base solution for the slow api is to introduce a background worker
- This can be kicked off manually via the info page (and maybe in the future run automatically - we have this flexibility)
- Suggestions and scores are stored in separate db fields, solely for the purpose of review
- To choose what suggestions to view, there can be a link in the info page saying "200 songs to review" which can then link to a modified bulk_updater page where songs can be bulk ticked/acccepted or edited one by one. This can be pretty flexible.
- The existing metadata_status field in the db tracks this process
- This system, due to the scoring,  could also allow a user to "automatically accept anything with a score over 100, or 120, or whatever.
- This overall process gives a fair bit of flexibility from fully automated to fully manual with the final decisions coming down to UX

PR #806, opened 2026-03-22
fix: prevent CJK characters from being stripped from YouTube filenames

safepay:
Emoji patterns have been adjusted to not strip CJK characters when cleaning YouTube file names.

### DIFF
--- a/pikaraoke/app.py
+++ b/pikaraoke/app.py
@@ -226,7 +226,7 @@ def main() -> None:
     app.config["SITE_NAME"] = "PiKaraoke"
 
     # Expose some functions to jinja templates
-    app.jinja_env.globals.update(filename_from_path=SongManager.filename_from_path)
+    app.jinja_env.globals.update(filename_from_path=k.song_manager.display_name_from_path)
     app.jinja_env.globals.update(url_escape=quote)
 
     spawn(upgrade_youtubedl)
--- a/pikaraoke/karaoke.py
+++ b/pikaraoke/karaoke.py
@@ -120,6 +120,7 @@ class Karaoke:
         show_splash_clock: bool | None = None,
         splash_delay: int | None = None,
         volume: float | None = None,
+        enable_title_tidy: bool | None = None,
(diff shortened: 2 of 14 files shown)


vicwomg:
The issue turned out to not be that of the arabic characters themselves, but TRAILING_NOISE_PATTERNS in metadata_parser.py was too broad—it stripped the word "Karaoke" and everything that followed it. This was the primary reason the Arabic translation in the filename was being scrubbed.

I am growing uneasy about this and the commit f71c14d68724caedc80373f6a641fe86aa25fb6f changes around cleaning yt filenames. Seems like there's a lot of potential corner cases here 

Why do we need to be so proactive about scrubbing this data? Seems like it should be opted in when the user decides to try to clean up their library themselves.

safepay:
> TRAILING_NOISE_PATTERNS in metadata_parser.py was too broad

Was already fixed in the feat/replace-lastfm-with-itunes branch along with a raft of changes.

safepay:
Would you like me to implement an admin switch to clean the YT file names? Or there another path you would like to take?

vicwomg:
> Would you like me to implement an admin switch to clean the YT file names? Or there another path you would like to take?

My thinking is that there's no need to clean the display filename until the user decides to rename the song via the edit or batch renamer. Some users might not care to do so and I think that's fine.

safepay:
I guess that for some people it might be a problem cleaning the YT file names on the fly, but for me it actually works really well and I would keep it as an option. The code has had a lot of reviews and there are very few edge cases - none that I can pick up for english content. Some languages might struggle, because the non-english karaoke songs can be labelled differently.

vic-resin:
I don't really want to compromise the experience for non-english language users

safepay:
> I don't really want to compromise the experience for non-english language users

It shouldn't be. I have spent a long time on this. It should be fine, but of course I can't test every single permutation and combination. If its optional, then it should benefit everyone except those with the edgiest of cases.

Issue #144, opened 2021-09-11
Select hat as input and have hdmi as output


sromo006:
I have a hifiberry hat and want to use that for input. I cant seem to get pikaraoke to get the input. What do i need to add when launching to get the input changed?

vicwomg:
Piping audio input to HDMI output is possible, as it is with a USB  microphone device, but my guess is that you're not going to find the latency usable

First find out the card/device number of your audio input, by running `arecord`

```
pi@pikaraoke:~ $ arecord -l
**** List of CAPTURE Hardware Devices ****
card 2: Device [USB PnP Audio Device], device 0: USB Audio [USB Audio]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
```

In this example, my USB microphone is Card 2 and Device 0. Therefore my device is "hw:2,0" which I plug into this arecord's -D parameter, and then pipe audio to playback. 

```
arecord -D hw:2,0 -f CD -c 1 -- | aplay
``

You'll be able to hear your mic through your TV, but it will be delayed by almost 1s. 

For now, this is why I do not recommend this sort of setup.

Issue #124, opened 2021-05-31
USB MIC not working


GST5555:
I conected an USB mic to the raspberry and the mic works fine on the raps, but with this application is useless, any idea how to mix the input of the mic with the karaoke ?

vicwomg:
You'll need to somehow pipe your USB audio mic input into the output (HDMI or line out). 
See: https://stackoverflow.com/questions/54452020/raspberry-pi-redirect-usb-microphone-to-hdmi-audio

I don't personally recommend this setup due to latency. You'll get 0.3-1 sec delay in audio output, and you will notice it. That's just the nature of USB audio. Instead, I recommend getting a separate set of microphones and a mixer. There is a link in my readme for an all-in-one solution (2x wireless mics with mixer built into the receiver)

vicwomg:
Piping audio input to HDMI output is possible with a USB  microphone device, but my guess is that you're not going to find the latency usable

First find out the card/device number of your audio input, by running `arecord`

```
pi@pikaraoke:~ $ arecord -l
**** List of CAPTURE Hardware Devices ****
card 2: Device [USB PnP Audio Device], device 0: USB Audio [USB Audio]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
```

In this example, my USB microphone is Card 2 and Device 0. Therefore my device is "hw:2,0" which I plug into this arecord's -D parameter, and then pipe audio to playback. 

```
arecord -D hw:2,0 -f CD -c 1 -- | aplay
```

You'll be able to hear your mic through your TV, but it will be delayed by almost 1s. 

For now, this is why I do not recommend this sort of setup. Unless someone can figure out how to get the latency down. But it seems to me like a limitation of the pi.

vicwomg:
fwiw, I did try tinyalsa as recommended in the above stackoverflow, and for hw device 2, the command was:

```
tinycap -- -D 2 -c 1 -r 48000 -b 16 | /usr/local/bin/tinyplay -
```

The latency was much better, but the playback sounded pitched up, like the chipmunks. No amount of messing with the sample rates seemed to fix it, but you may have better luck with your setup.

PR #818, opened 2026-04-14
feat: experimental microphone passthrough

safepay:
**_Very experimental and needs testing help_**

Adds server-side mic passthrough so singers can hear themselves through karaoke speakers.

**What was tried:** Piping raw audio via `arecord | aplay` (ALSA) produced ~300–1000ms latency; `tinyalsa` (`tinycap | tinyplay`) had better latency but pitch-shifted audio upward with no clear fix.

**This approach:** Uses the audio system's own loopback instead of piping audio through user-space:

- **Linux (PulseAudio/PipeWire):** `pactl load-module module-loopback` routes mic to speakers entirely within the audio server — no pitch artifacts, configurable latency (10–200ms), optional WebRTC echo cancellation.
- **Windows/macOS:** `sounddevice` (PortAudio) duplex RawStream callback with mono-to-stereo upmixing.

Mic devices are listed on the `/info` page with per-device enable/volume controls. Settings persist across restarts. The installer auto-adds `pulseaudio-utils` on Linux if missing.

> **Experimental** — latency and echo cancellation effectiveness will vary by hardware and audio server version. Real-device testing needed.

Attempts to address:
- #144 
- #124 
- #797 
- #275 
- #242


prairiewest:
I tested this today. It did not work for me at first, but I found the problem: pactl needs the environment variable XDG_RUNTIME_DIR to be set or it will throw an error.  In my case the value needs to be "/run/user/1000", because I'm running pikaraoke as user uid=1000.

The reason that this env var is missing for me at run time is that I am starting pikaraoke using systemctl.  So I can add the env var to my service file, however I think it wouldn't be obvious to other people that this env var is needed.  Alternatively, one can add "env=" to each subprocess.run call, but it needs to be "/run/user/[UID]" where UID is of the user that starts the pikaraoke process.

I'll do some more testing tomorrow regarding the actual sound quality, comparing the mic audio through the USB adapter and then into server loopback versus microphone straight into the amplifier mic input.

prairiewest:
After comparing passthrough audio with direct-to-amplifier audio, I wouldn't use passthrough for two reasons:
- there was a noticeable delay (50 ms? not sure, I don't have equipment to measure it)
- there was clipping of the microphone audio signal

I don't know if all of this was caused by my cheap USB microphone adapter? Perhaps it was. Here is the one I used: https://www.amazon.com/dp/B01N905VOY/

prairiewest:
Now that I look at your code I see that _MAX_GAIN is set to 2.0. I saw that the volume slider in the Pikaraoke GUI was about half way, which I assumed would be at 50% volume, however that was 100% (half of 2.0).  I'm not sure that _MAX_GAIN of 2.0 is a good default, I would use 1.0.

I changed _MAX_GAIN to 1.0, _DEFAULT_LATENCY_MS to 10, and updated the GUI templates locally.  After restart and setting the mic volume to 60% and latency to 10ms, it's all working better now.

I think this code could be merged if it was enabled only with a feature flag, such as: `--enable-mic-passthrough`

safepay:
> Now that I look at your code I see that _MAX_GAIN is set to 2.0. I saw that the volume slider in the Pikaraoke GUI was about half way, which I assumed would be at 50% volume, however that was 100% (half of 2.0). I'm not sure that _MAX_GAIN of 2.0 is a good default, I would use 1.0.
> 
> I changed _MAX_GAIN to 1.0, _DEFAULT_LATENCY_MS to 10, and updated the GUI templates locally. After restart and setting the mic volume to 60% and latency to 10ms, it's all working better now.
> 
> I think this code could be merged if it was enabled only with a feature flag, such as: `--enable-mic-passthrough`

Thanks for that feedback and for testing. I've made those changes. If you wouldn't mind retesting. It is essentially the same code with different defaults and the startup flag.

prairiewest:
No problem. I re-tested on a clean pull of the updated code, it worked well.  I tested with the new feature flag, and the feature was enabled.  Tested without the feature flag, and the feature was gone and I didn't see any problems or leftover GUI or anything like that.

safepay:
> No problem. I re-tested on a clean pull of the updated code, it worked well. I tested with the new feature flag, and the feature was enabled. Tested without the feature flag, and the feature was gone and I didn't see any problems or leftover GUI or anything like that.

@prairiewest thanks for testing. I've made those changes and switched the PR to ready-to-merge.

toasterboy1:
Experimental feature does not seem to work on my MX-Linux setup on a very old laptop (Thinkpad E530) with cheap USB wireless Microphone. I ended up just using a bash script to toggle pw-loopback on before I start and then to toggle it off when I am done. There still seems to be a bit of delay but not really sure how bad it is. Just running sound to my TV via HDMI, nothing professional. Not sure if this would be helpful for others or with the Raspberry Pi.

PR #819, opened 2026-04-15
feat: add subpath support for reverse proxy deployments

### LINKED ISSUE #812: [FEATURE] Run behind a reverse proxy in a subpath
**Is your feature request related to a problem? Please describe.**
I really like that Pikaraoke can run headless on a server and be controlled from any browser. However, when exposing it to the public internet, some form of protection is needed.

In my setup, the simplest approach is to run the app behind a reverse proxy under a long, non-guessable subpath (e.g. `/some-random-path/`). Currently, Pikaraoke assumes it is served at `/`, and various links and asset paths break when using a subpath.

**Describe the solution you'd like**
Make the base path (subpath) configurable.

I explored this fairly quickly and ended up with a working implementation in my fork:
https://github.com/childerich/pikaraoke/tree/fix/subpath-support

It touches ~10 files (mainly around path handling), so before polishing it further I wanted to check whether this is something you'd consider in scope.

**Describe alternatives you've considered**
- Basic Auth via reverse proxy, but that can be inconvenient on some client devices
- A separate subdomain, which works for deployment but does not address my goal of hiding the app behind a non-guessable URL

**Additional context**
My fork works fine in my setup (Docker + nginx-proxy). However, maintaining a fork long-term is not ideal, especially with frequent upstream changes (e.g. yt-dlp updates).

I understand this might be a niche use case, so I wanted to check first before investing more effort into a PR.

childerich:
## Add Subpath Support for Reverse Proxy Deployments

**Problem**: PiKaraoke assumes it runs at the root path (`/`), preventing deployment under subpaths like `/karaoke` for reverse proxy setups.

**Solution**: Add `--base-path` CLI argument to serve PiKaraoke under any URL prefix while maintaining full backward compatibility.

### Changes
- New `lib/url_prefix.py` middleware for path rewriting
- Updated path handling in templates, JavaScript, and Python code
- Added tests for URL prefix functionality
- Updated documentation with usage examples

### Usage
```bash
pikaraoke --base-path /karaoke
```
Addresses [issue #812](https://github.com/vicwomg/pikaraoke/issues/812)


vicwomg:
Looks good at a glance! I will have more time to thoroughly review later this week

vicwomg:
Is it just me or does the splash screen continually throw this error?

Screenshot 2026-05-03 at 5 19 18 PM

Looks like a 404 error trying to connect to the socket:

```
Failed to load resource: the server responded with a status of 404 (NOT FOUND)Understand this error
splash.js:567 Connection error: n: xhr poll error
    at i.onError (socket.io-4.8.3.min.js:6:8567)
    at n.<anonymous> (socket.io-4.8.3.min.js:6:12051)
    at I.emit (socket.io-4.8.3.min.js:6:7218)
    at i.B (socket.io-4.8.3.min.js:6:13505)
    at socket.io-4.8.3.min.js:6:13277
```

`http://192.168.86.XXX:5555/karaoke/socket.io/?EIO=4&transport=polling&t=ge3nz8zm`

childerich:
I was able to reproduce this when running in local docker with subpath but without reverse proxy. Sorry I'll work on a fix

childerich:
I tinkered around a while with this until i realized its not possible to support that config value properly and stable in all different environments. I now only extended the documentation to make clear that this feature is only intended to work with reverse proxy setups. Hope that makes sense

vicwomg:
> I tinkered around a while with this until i realized its not possible to support that config value properly and stable in all different environments. I now only extended the documentation to make clear that this feature is only intended to work with reverse proxy setups. Hope that makes sense

My concern is that it doesn't work even in the most basic setup: no container, just running locally with `uv run pikaraoke --base-path /karaoke` , if that doesn't work then the feature seems like it is intended for very niche configurations.

childerich:
Good that you questioned this... I had tried this fix before but it caused problems in my reverse proxy setup - but i just got it running with a fresh local setup. Not sure where the problem was (potentially only in my head).

safepay:
Heads up on an unrelated typo in queue.html that I think slipped in with this change. In executeQueueAction, the highlight-tracking line was renamed:

```
// Track moved file for highlight (except delete)
if (action !== 'delete') {
    window.highlightAfterHighlight = file;   // <-- should be highlightAfterRefresh
}
```

childerich:
Thanks a lot @safepay. Fixed it and found a few more places that needed the subpath support

vicwomg:
Works well on localhost now!

vicwomg:
This is solid. thank you for your patience seeing this PR through.

childerich:
Thank you for integrating this into your great project!

PR #823, opened 2026-04-29
feat(lyrics): mobile lyrics panel on the now-playing view

zygzagZ:
## Summary

- **New mobile lyrics panel** on the expanded now-playing view: scrollable lyric lines with the current line highlighted in amber and auto-centered. Backed by the existing `.ass` file served at `/subtitle/<id>` and the `playback_position` socket events already on the wire — no backend changes, no new endpoints.
- **Pure ASS parser ES module** (`lyrics-parse.js`) with 22 unit tests via vitest. Handles PiKaraoke's rolling-window ASS format (extracts the `\b1` "current" segment, dedupes context lines) and pulse-tagged `\kf` chunks (regex allows `\kf` anywhere in the override block, so the `\t(...)` injected by `lib/lyrics.py:_k_token` doesn't drop the first character of every word).
- **Word-level fill via rAF** capped at ~16 fps using `background-clip: text` driven by `--pk-fill-pct`. The rAF lifecycle deliberately ignores `data.is_paused` — server's `is_paused` can lag the splash's actual state, so we use socket-position recency (>1.8 s of quiet = pause) instead.
- **Tap-to-seek** on lyric lines (admin-only), keyboard accessible (Enter/Space), with WCAG-compliant 44 px toggle target.
- **Live re-sync** on `preferences_update` (subtitle-offset slider) without waiting for the next 1 Hz position tick.
- Composes cleanly with the subtitle source picker (`?v=` bumps trigger abort + refetch + rebuild; `subtitle_source_override === 'off'` clears the panel).

## Test Coverage

```
[+] pikaraoke/static/js/lyrics-parse.js (NEW, pure parsers)
    parseAss / parseAssTime / findActiveLineIdx
    └── 22 vitest tests cover Comment skip, override blocks,
        rolling-window hot-segment extraction, pulse-tagged \kf
        chunks, \N break, empty-after-strip, Polish characters,
        comma preservation, sort order, boundary/gap/latch.
        [★★★ TESTED]

[+] pikaraoke/static/js/now-playing-bar.js (DOM/socket glue, ~360 added)
    updateLyrics, buildLyricsDom, tickLyrics, setActiveLine,
    scrollActiveIntoView, rAF word-fill, tap-to-seek, toggle,
    preferences_update listener
    └── Verified end-to-end on a real phone during dogfooding.
        Three bugs surfaced and fixed during that loop:
        - rolling-window dedupe (each line was rendering 5×)
        - pulse-tag regex (first char of every word was dropped)
        - is_paused over-gating (smooth fill stopped while audio
          was still playing because server is_paused was stale)
```

Tests: **22 / 22 vitest passing** + **1522 / 1522 pytest passing** (no Python touched, sanity check only).

## Pre-Landing Review

- XSS: `textContent` for all lyric text; querySelector idx is internal-only — clean.
- Memory: rAF cancelled on stop + auto-exits when position events go stale; AbortController on URL change; scroll/raf timers cleared on close & clear.
- Race conditions: AbortController + URL-staleness check inside `.then()` + AbortError swallow; rAF singleton via `state.rafId`.
- Cross-browser: `matchMedia.addEventListener`/`addListener` fallback; `scrollend` feature-detected; `-webkit-background-clip` paired with standard.
- Accessibility: `aria-live="off"` (intentional — scrolling lyrics would spam screen readers), `role="button"` + `tabIndex` on tappable lines, `aria-label` on toggle, full `prefers-reduced-motion` overrides.
- [INFO, non-blocking] If `parseAss` ever returns `[]` for a valid-but-empty ASS file, the next `now_playing` event would re-fetch the same URL. Low severity in practice (URL changes on every regen via `?v=`); flagged as a follow-up nit.

## Test plan

- [x] Vitest passes locally (22 / 22)
- [x] Pytest passes locally (1522 / 1522, sanity)
- [ ] On a phone, confirm: cover shrinks to thumbnail, lyrics panel appears, current line is amber-highlighted and auto-centers
- [ ] Manual scroll mid-song pauses auto-scroll for ~3 s, then resumes on the next line change
- [ ] Subtitle-offset slider drag re-syncs the highlight immediately
- [ ] Hit skip/next: lyrics clear and re-populate, or panel hides if next song has no lyrics
- [ ] As admin, tap a lyric line — splash seeks to that line's start time
- [ ] Tab through lines with a keyboard — Enter/Space seeks
- [ ] Toggle the collapse button — cover restores, lyrics state preserved on re-expand
- [ ] Subtitle source picker: pick `off` clears the panel; pick a different source rebuilds from the new variant


zygzagZ:
Wrong target — re-opening in fork.

zygzagZ:
sorry about that, the agent used wrong target

PR #835, opened 2026-07-04
feat: keep awake headless

safepay:
Add --keep-awake flag

Prevents the host from sleeping while PiKaraoke runs (Windows execution state, macOS caffeinate, Linux systemd-inhibit). Useful in headless mode, where there's no local player window to keep the machine awake.


safepay:
Note that the system's attempt to upgrade yt-dlp is breaking the smoke-test on Mac.

vicwomg:
> Note that the system's attempt to upgrade yt-dlp is breaking the smoke-test on Mac.

Experimenting with some more debug output on https://github.com/vicwomg/pikaraoke/pull/838

Looks like the actual error is:
`[2026-07-07 18:00:24] INFO: ERROR: Unable to obtain version info (HTTP Error 403: rate limit exceeded); Please try again later or visit `

Quite strange that it only occurs on the macos runner.

vicwomg:
> > Note that the system's attempt to upgrade yt-dlp is breaking the smoke-test on Mac.
> 
> Experimenting with some more debug output on #838
> 
> Looks like the actual error is: `[2026-07-07 18:00:24] INFO: ERROR: Unable to obtain version info (HTTP Error 403: rate limit exceeded); Please try again later or visit `
> 
> Quite strange that it only occurs on the macos runner.

For now I have removed the osx smoke test runner. Ubuntu seems to cover the bases for unix-like OS

safepay:
yt-dlp -U self-updates by first hitting GitHub's API for version info, and GitHub's macOS runners share a small pool of outbound IPs that blow through the unauthenticated 60-req/hr limit — hence the macOS-only 403; since we always pip-install yt-dlp anyway, the cleanest fix is a --skip-ytdl-update flag that the smoke test passes so CI never makes that call.

vicwomg (approved):

Issue #90, opened 2021-01-02
Singer / Song Play history


WhatWhyIT:
Ability to keep / set how many days to keep Singer / Song history.

Maybe default  to 1 day.

As an added enhancement to this request - have ability to keep a running total for how many songs a singer has sung. 👍

safepay:
Resolved in #880 to comply with the majority of user's requests.

Open another issue if there is something needed specifically.

PR #845, opened 2026-07-19
feat: Session Management and Play History

### LINKED ISSUE #213: Feature request: generate playlist of all played songs in one sitting
We'd like to have a way to generate playlist of all played songs in one sitting so maybe file per day 
with interpret and song name or with the file name

safepay:
Implements the v1 PR described in #793: a passive record of who sang what and when, built on the existing song database.

### What it adds

- **`sessions` + `plays` tables** on the existing SQLite DB
- **Sessions** — named time brackets. Name, stop, reactivate, or delete from the Sessions page.
- **Session Name** — Appears on the app pages to let everyone know which session is active
- **Show/Hide controls** — New Settings options to show/hide the logo and/or the session name on the splash screen
- **Automatic play recording** — `PlayHistoryManager` logs each play once playback succeeds.
- **Sessions page** (admin) — run the night: name, stop, reactivate or delete a session, with a per-session singer list.
- **Play History page** (open to the room) — the paginated log, sortable on every column and filterable by session, performer or song. Performer names and song titles are links that filter the log to their own plays, so a singer can pull up what they sang and re-queue it. Deleting an entry stays admin-only.
- **Rankings page** (open to the room) — top songs, top performers and busiest sessions, all-time or scoped to one session, with row-count controls. Chart rows link through to the plays behind them.
- **Export** — a session's log as CSV or a numbered, human-readable text list (satisfies #213).

Deferred to follow-ups per #793: a dedicated performers page, performer aliases, and retention. The schema supports these without further changes.

Split into #857: **KJ mode**, the opt-in `?kj_mode=true` singer field for a host queuing on someone's behalf, so it can be reviewed and released as its own line item. Queue attribution here is the device's own name, as before.

### Issues resolved

Closes #793.
Closes #213 — a session's log exports as a numbered, human-readable set list.

Two others are partly addressed but should stay open:

- **#90** — the running total per singer is delivered by the rankings, but the primary ask, configurable retention, is still deferred.
- **#209** is a discussion rather than an issue, so no keyword applies. Its request is met: performer and song links in the log let a singer find what they sang, across sessions, and re-queue it.

### Implementation notes

- **The active session is cached in `PlayHistoryManager`.** The nav ribbon reads it on every page render and the `now_playing` broadcast on every queue or playback change, so an uncached read meant a SQLite query — under the shared connection lock — on both hot paths. The manager is the only writer to the `sessions` table, so all five mutations route through one `_announce_session_change()` that drops the cache and emits `session_changed`.
- **Singer attribution resolves in one place.** `resolveSinger()` in `base.html` is the single rule (device cookie, prompting once if unset); the search and files pages both call it rather than each carrying their own variant.
- **The Sessions page unbinds its socket handler on SPA navigation away.** `window.socket` deliberately survives page swaps, so a handler left bound would keep re-querying the play log for a table no longer in the DOM.

### Test plan

- [x] Finish a song → appears in `/sessions`, marked Played; skip one → Skipped.
- [x] Rename, end, reactivate, delete a session; counts, log, and the session-name ribbon all update.
- [x] Filter by session, sort by time/performer/song, page both lists.
- [x] As a non-admin: `/history` and `/rankings` both open, with no delete controls; `/sessions` redirects home.
- [x] Click a performer name, then a song title, in the log — each filters to that performer's or that song's plays; re-queue from there.
- [x] Scope the rankings to a single session, then back to all time — counts change accordingly, and a chart row links through to the matching plays.
- [x] Filter by a session, then page the sessions table until that session scrolls off — the singer list heading still names the filtered session.
- [x] Export a session as CSV and as text; text is a clean numbered set list.
- [x] `/rankings` shows top songs/performers; row-count control works and each list's "N plays" column reads correctly.
- [x] Queue a song from **both** the search page and the files page — each is attributed to the device name.
- [x] Splash screen shows the session name; toggling the Settings hide/show controls for logo and session name takes effect.
- [x] Open `/sessions`, navigate away to `/search`, then play a song — no background requests to the history endpoints in the network tab.


vicwomg:
Seems like there's still activity here. Marking as draft PR. Feel free to move it back into "ready" when it is

safepay:
It's ready to go now. I'll just fix the merge conflicts

safepay:
There will be still some tweaking from feedback, so I would suggest doing those in a follow-up PR, rather than continuing to tweak this one.

safepay:
One thing that needs resolution - the KJ mode. It could be a decision that this is NEVER implemented and that PiKaraoke stays  social only. The KJ mode essentially starts to diverge the code base and creates a problem for new features.

It can be handled, but it just complicates future features.

It could also be a CLI arg, rather than a query arg.

safepay:
I've set this to draft because it is a major feature and requires feedback before finalising the PR, although I am now happy with it, following a bit of a UI rethink.

When you have time, please review and leave note any changes/issues.

vicwomg:
Sorry took me a while to look at this! Looking good, some feedback after playing with it a bit.

- Session name bar on  all screens margins don't align with the other containers, its slightly shorter in width. Containers should match. 
- I think sessions and play history are different things and it might be good to have a separate page for each. The play history page could have a dropdown of the current/past sessions that serve as the filter, which defaults to "ALL"
- Play History: I'm finding the mobile table revised layout/sorting buttons to be unconventional. Can we keep it columns? We can save some horiz space by collapsing the date to a smaller format on mobile / removing second-precision 8/1/26 16:42, but really not concerned with crowding overall as long as word wrapping is working.
- Play History: It seems like one of the main reasons I'd want to look at play history is to re-add songs I've sung before. Maybe we should have a settings icon and stash an "add to queue" and "delete" in a resulting modal. 
- Rankings: Top songs / Top performers: column label headers would be good (rank, song, plays), so we don't have redundant text like" 1 play, 2 plays
- Rankings: having the session title inline with the "all time" / "this session" selector causes a weird line break to the selector control on mobile. Maybe put the title on the next row. Or we can has a dropdown that will Filter by: ALL and the individual sessions in the db.
- Top songs: top songs should be queue-able to current queue
- Tables in general: can we just agree upon a reasonable per-page like 50 and not make it configurable? I like the new paging controls (1-10 of 50) , but they don't match that on the browse screen. Maybe we should update the browse screen to match.
- Play history: I think show skips should be default and "hide skips" should be the checkbox.
- Session history: Same here, show all by default and the checkbox will "Hide auto-started sessions"
- Session history: "Current Session" name has unnecessary horizontal margins and doesn't align with the rest of the page. Lets remove those. 
- "Reset all history" should be in general settings

safepay:
Thanks! All feedback taken on board.

The main things to note are now that:

* Users can see Rankings and Play History and requeue songs.
* Sessions and Settings are is_admin gated
* The pagination has COMPLETELY CHANGED! This is now simplifed and unified across browse as well. I think this will fit nicely in to PR #853 as the song list will generally be smaller once someone searches.

Let me know your thoughts.

vicwomg:
Looks great. A couple of very small tweaks and I think this is ready if you are
- For dropdowns use the bulma styled dropdown: https://bulma.io/documentation/components/dropdown/
- Can sessions be grouped above the separator line with rankings and play history? Seems out of place. I think settings should be isolated from those features

Screenshot 2026-08-12 at 2 22 31 PM

safepay:
I'll look at the Bulma drop-down.

Regarding the menu, I am trying to distinguish admin-exposed functions from all-user functions.
Settings and Sessions are admin only, while Rankings and History are visible to everyone.

Not sure I have a better solution than the two below the line.

vicwomg:
> I'll look at the Bulma drop-down.
> 
> Regarding the menu, I am trying to distinguish admin-exposed functions from all-user functions. Settings and Sessions are admin only, while Rankings and History are visible to everyone.
> 
> Not sure I have a better solution than the two below the line.

Ok then put settings last. It has more "gravity" so to speak

safepay:
And I want to split the ?kj_mode=true functions in to a separate PR for visibility and review.

safepay:
Note that this pages already use Bulma's `navbar-dropdown` as documented here https://bulma.io/documentation/components/navbar/ instead of in the other link.
The one being used is better at supporting mobile screens.

PR #849, opened 2026-07-21
ci: re-enable macos latest smoke test

vicwomg:


safepay:
Changing the download worker to a Gevent greenlet did not fix the problem, but it's worth keeping. It was the reverse-DNS that actually fixed it.

vicwomg:
thanks for the assist @safepay

PR #853, opened 2026-07-26
feat: separate song library search from YouTube search

### LINKED ISSUE #146: Auto-completion prevents from putting cursor at the end of the search box
I'm on android phone with a tactile keyboard. 
You can only delete backward so sometimes you need to put the cursor at the end of the search box to erase. 
But clicking on the right of the searchbox trigger auto-completion and the cursor stay in place.

### LINKED ISSUE #851: [FEATURE] Separate the local library from YouTube search
New users can't tell **Browse** and **Search** apart — the names are near-synonyms, and the
pages overlap: the Search box runs a local-library autocomplete *and* a YouTube search, doing two
unrelated jobs at once. Meanwhile Browse has no way to find a song by name, only an A–Z bar.

Give each page one job:

- **Browse → "Songs"** — songs already on this machine. Gains a filter-as-you-type box.
- **Search → "Add New"** — find and download new songs. Loses all local-library functionality
  (the autocomplete is removed).
- A YouTube result that's *already downloaded* shows an "add to queue" action instead of a
  pointless second download.

Nav labels and page headings get renamed to make the local/internet distinction obvious; routes
(`/browse`, `/search`) stay the same so bookmarks don't break.

## Notes

- Filtering runs server-side (Browse is paginated, so a client-side filter would miss songs on
  other pages) and matches the raw filename, so words like "karaoke" that get tidied out of the
  display name are still findable.
- Perf-sensitive on Raspberry Pi — the current alpha bar already sweeps the whole library on every
  tap, which this replaces with a cached index.

safepay:
This is a fairly big UI change to hopefully improve usability around both searching the existing song library and adding new songs via YouTube. I found that guests simply "did not know what to do", so I hope this addresses that concern.

### What's changed

**The Songs page (was "Browse") has a filter box.** Type any part of a title or artist and the list narrows as you go. Filtering runs on the server against the whole library, so it finds songs on page 47 — not just the ~100 rows currently rendered. Multiple words all have to match, accents are ignored (`bjork` finds `Björk`), and the filter survives switching between alphabetical and date sort.

**It stays responsive on a Pi.** Each keystroke re-renders only the results table, not the whole page, and matching runs off a cached filename index rebuilt only when the library changes. Single Latin characters don't trigger a request at all — they match nearly everything, so it's maximum work for no narrowing. Single CJK characters do, since one character is often a whole word.

**The Add New page (was "Search") flags songs you already have.** A YouTube result that's already in your library now offers "Add to queue" instead of inviting you to download it a second time.

**Add New rows now follow the download instead of guessing at it.** Requesting a song used to flash a green tick and then revert to the original button after three seconds. But the request only hands the song to the serial download worker, so the tick claimed it had arrived before yt-dlp had started, and the restored button invited a duplicate request for a file already on its way. A row now holds an "Adding..." state until the worker reports the file, then becomes "Add to queue" with an "In library" badge — the same state the server renders for a song already held. That arrives over the socket, so every open browser updates, not just the one that clicked. A download that fails hands the choice back rather than leaving the row stuck.

**Sorting moved out of the way.** Alphabetical/Date is now a small toggle sitting directly above the table it sorts, showing both options at once, so the current ordering is visible at a glance.

**Assorted UI fixes:**
- Pasting a YouTube URL into Add New no longer gets blocked
- Search options right-align properly on mobile
- Per-row download actions on Add New
- Searching no longer leaves the mobile keyboard covering the results it just produced; focus returns via Clear, a fresh page load, or tapping the box
- On the Songs page the row number now sits on the line of the title it numbers, rather than 3px above it
- Tighter navbar icon-to-label spacing; nav icons now carry accessible labels (they're the only signal below 1088px)

### Note
- UI labels changed but translation catalogs are unchanged; a follow-up PR is needed to regenerate them.

### Test plan

- [x] **Songs page** — type a partial title; the list narrows. Clear it; everything returns.
- [x] Filter to a set with more than one page, click page 2 — results stay filtered.
- [x] Type quickly and backspace mid-request — the table always matches what's in the box.
- [x] With a filter active, click "Sort by Date" — the filter carries over. Click a letter in the A–Z bar — the filter clears.
- [x] With a filter active, click "Songs" in the navbar or the "show all" icon in the A–Z bar — the full list returns. Press Back — no full page reload.
- [x] Check the row numbers line up with their titles, including a title long enough to wrap onto two lines.
- [x] **Add New** — search for something already in your library; that row offers "Add to queue". Confirm it queues.
- [x] Click "Save for later" on a song you don't have — the row shows "Adding..." and both buttons go away. When the download finishes it becomes "Add to queue" with an "In library" badge, with no reload.
- [x] Do the same with "Add to queue" — the song is queued automatically and the row still ends up in the library state.
- [x] Leave Add New open on a second device while a download runs — the matching row there updates too.
- [x] Paste a YouTube URL into Add New and confirm the download starts.
- [x] On a phone, run a search — the keyboard drops rather than covering the results. Tap "Clear" — it comes back.

Closes #851
Closes #146


vicwomg:
I think this makes sense, but please adjust the language to "Add new songs from YouTube" and avoid the "d" word

safepay:
Done. Good to go. Except for a couple of UI refinements.

hayleox:
Ah, this PR has so many improvements that I was considering making my own fork for! Thank you! I'm running this locally now and it seems to be working flawlessly as far as I've found.

One UI tweak I'd suggest – in the navbar, I would put "Songs" before "Add New". I would also love if "Add New" could be changed to something that makes it more explicit that you're leaving the curated library and venturing into the wild west of YouTube. Maybe "Add from YouTube", or "Search YouTube", or even just "YouTube"?

safepay:
@hayleox I agree. I have swapped the Songs and Add New.

However, to keep an already large number of UI changes consistent for the user base I opted for "Add New" because that is the text on the original page (with "Search /" in front).

Your other naming suggestions were considered, and are still on the table.

safepay:
The last change shows the A-Z always on mobile. Users found it hard to know what to do on mobile in terms of searching by letter, because the A-Z was invisible until scrolling. This change makes that always visible in a way that does not interfere with the other page elements.

And ported the simplified paging from the Play History branch as this fits better with the new searching capabilities.

Now its good to go!!!

Issue #858, opened 2026-08-12
FFmpeg 9.0 removes -vsync, breaks all local video/YouTube playback on Windows


D4RKM47R1X:
Describe the bug

Every song fails to play with FFmpeg exited with code 2880417800 immediately, before any transcoding starts. Reproduced with 3 unrelated local library files (different filenames, different sources), so it isn't file-specific.

Steps to reproduce
Install PiKaraoke 1.21.0 on Windows with FFmpeg 9.0 on PATH.
Add any video to the song library (or download one from YouTube).
Play the song.
Playback fails immediately.
Debug log (-l 10)
COMMAND: ffmpeg -i <file> -map 0:a -map 0:v -f hls -b:v 5M -b:a 192k -ac 2 -acodec aac -ar 48000 -avoid_negative_ts make_zero -hls_fmp4_init_filename ... -hls_list_size 0 -hls_playlist_type event -hls_segment_filename ... -hls_segment_type fmp4 -hls_time 3 -preset ultrafast -vcodec copy -vsync cfr <output>.m3u8
[FFMPEG] Unrecognized option 'vsync'.
[FFMPEG] Error splitting the argument list: Option not found
ERROR: FFmpeg exited with code 2880417800
ERROR: Failed to prepare stream

Environment
PiKaraoke version: 1.21.0
OS: Windows 10.0.26100 (Windows 11 24H2), amd64
FFmpeg: 9.0-full_build-[www.gyan.dev](http://www.gyan.dev/)

Root cause
FFmpeg 9.0's release notes confirm -vsync was fully removed from the CLI (replaced by -fps_mode), after years as a deprecated-but-working alias: https://jbkempf.com/blog/2026/ffmpeg-9.0/ — "On the command line, -vsync is finally removed (use -fps_mode)..."

PiKaraoke hardcodes the old option in pikaraoke/lib/ffmpeg.py (line ~165):  "vsync": "cfr",

vicwomg:
Oh thats interesting. Do you happen to know if the install scripts auto-install ffmpeg 9.0 now, or did you just have that version installed system wide separately?

D4RKM47R1X:
I had an older version so I think it does but it only installs if ffmpeg isn't already found on PATH, so it probably never upgrades an existing install.

PR #861, opened 2026-08-12
fix: replace deprecated -vsync with -fps_mode for ffmpeg 9.0 compatibility

D4RKM47R1X:
Fixes issue #858


vicwomg:
is this backwards compatible with users on older ffmpeg?

D4RKM47R1X:
It is backwards compatible up to ffmpeg 5.1 so you could add a minimal requirement or have it detect version and then decide which to use.

vicwomg:
Great, seems fine. I think we generally recommend v6

PR #873, opened 2026-08-18
fix: rank the fair queue by turns taken this session

### LINKED ISSUE #870: [BUG]  Fair queue ignores the currently playing singer when calculating   rounds
**Describe the bug**
When fair queue is enabled, the round-robin calculation only considers songs that remain in the queue. It does not account for the singer whose song is currently playing.
See example below, seems minimal, but bothered more than once and I think its a simple fix. (also, I gave a suggestion on replacing round robin with an actual spreader algo)

**To Reproduce**
(remember the queue order from A,B,C, it will matter for the last section)
  1. Enable **Fair queue**.
  2. User A queues five songs: `A1`, `A2`, `A3`, `A4`, and `A5`.
  3. Start playing `A1`.
  4. While `A1` is playing, user B queues `B1`.
  5. User C then queues `C1`.
  6. Observe the resulting queue.

**Expected behavior**
From the reproducer steps, what I expect would be: **A1, B1,** C1, A2, A3, A4, A5

What actually happens:  **A1, A2,** B1, C1, A3, A4, A5

**Platform (please complete the following information):**
i use arch btw (I think error is not related to plataform)

**Additional context** -> actual spreading algorithm! (I can open a PR if this feat make sense, it is what I have been using locally for a while)
 An alternative policy would be to spread a singer's existing backlog across newly arriving singers instead of completing a strict round first!! See, from the example above, the queue could be, instead:

**A1, B1, A2, C1**, A3, A4, A5

We put C1 after A2, so A get more interleaved!, that would incentivize singers to batch add songs they want to sign, and their performances would be spaced the best way possible.

safepay:
## Fair queue: rank by turns taken, not queue position

Closes #870

The round-robin position was worked out from what each singer had left in the queue, so any turn already consumed was invisible to it. The song on screen has left the queue, which is why a newcomer's first song landed *behind* the second song of the singer at the microphone. The same blind spot has a larger form: when the queue drains, everyone resets to equal no matter who has sung eight songs.

Rounds are now ranked by how many songs each singer has sung tonight, read from play history:

- **Skipped songs are not turns.** A song pulled after ten seconds keeps the singer's place rather than sending them to the back for something they never sang.
- **The song on screen counts**, taken from the now-playing state — its play row stays unresolved until the song ends, so the two sources can't double-count it.
- **Catching up is capped at the round that just played.** An under-served singer always goes ahead of a well-served one, but the head start is one round. Without the cap, someone arriving at midnight with no turns would claim every round the room got through before they walked in.

Scoped to the active session: a name that sang ten songs last Friday walks in tonight owing nothing.

The issue also suggests replacing rounds with a spreading algorithm on the grounds that it would encourage batch queuing. That is left out deliberately — rewarding whoever queues most aggressively runs against what the setting is for, and the starvation it targets is the latecomer case, which the cap already covers.

### Test plan

With **Fair queue** enabled:

- [x] A queues five songs; while A's first plays, B then C each queue one. Order becomes A1, B1, C1, A2 — not A1, A2, B1.
- [x] A sings two songs and the queue empties. A and B both queue one: B's plays first.
- [x] Skip B's song a few seconds in. B queues again and keeps their place instead of dropping behind.
- [x] After several rounds, a new singer queues three songs: they get one turn at the front, then interleave normally.
- [x] Turn fair queue off: songs append in the order they were added, as before.

### DIFF
--- a/pikaraoke/karaoke.py
+++ b/pikaraoke/karaoke.py
@@ -295,6 +295,8 @@ class Karaoke:
             get_now_playing_user=lambda: self.playback_controller.now_playing_user,
             filename_from_path=self.song_manager.display_name_from_path,
             get_available_songs=lambda: self.song_manager.songs,
+            # Bound late: play history is built below.
+            get_turns_taken=lambda: self.play_history.get_turns_taken(),
         )
 
         # Play history recording and reporting (subscribes to song_ended itself)
--- a/pikaraoke/lib/play_history_manager.py
+++ b/pikaraoke/lib/play_history_manager.py
@@ -624,6 +624,27 @@ class PlayHistoryManager:
         )
         return [dict(row) for row in rows]
 
+    def get_turns_taken(self) -> dict[str, int]:
+        """Return how many songs each singer has sung through tonight, keyed lower-case.
+
+        What the fair queue shares turns out against, so it is scoped to the
+        active session and counts neither skipped songs nor the one on screen,
+        whose row does not settle until it ends.
+        """
+        session = self.get_current_session()
+        if session is None:
+            return {}
+        rows = self.db.query(
+            f"SELECT p.performer, COUNT(*) AS turns FROM plays p "
+            f"WHERE p.session_id = ? AND {_SUNG} GROUP BY p.performer COLLATE NOCASE",
+            (session["id"],),
+        )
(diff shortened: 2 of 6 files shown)


rodrigo-ceccato:
I will run this branch locally from now on, if anything comes up I will bring updates... But firing it up locally looked nice so far o/

Issue #877, opened 2026-08-19
[BUG] Adding songs from the songs browsers always shows "Song is already in the queue: <song title>" toast


vicwomg:
**Describe the bug**
Adding songs from the songs browsers always shows "Song is already in the queue: <song title>" toast

**To Reproduce**
1. Go to song browser
2. Click + icon next to any song that is not in the queue already
3. Observe "Song is already in the queue" toast

**Expected behavior**
"Song is already in the queue" toast should only display if the song is already in the queue. For first time add, it should not show

**Screenshots**
N/A

**Platform (please complete the following information):**
 - Hardware: Macbook
 - OS: OSX
 - Browser: Chrome

**Additional context**
Seems new to recent changes

vicwomg:
@safepay not sure if this is the result of recent changes but saw it QAing today

safepay:
@vicwomg, as you have a Mac, can you reproduce it?

After a full code audit and testing, I can't reproduce this.

safepay:
However there is a small bug on the new Rankings page, which I will ship a fix for.

vicwomg:
I filed this issue.

Yes I can reproduce it readily. Testing further, it seems to occur on localhost:5555, but not when using the full IP URL. Can you try localhost:5555?

safepay:
Tried it. No issues at all.

I am running one version on Windows and one in Docker/Linux and cannot reproduce that.

safepay:
I just checked and none of the code touching this has been changed since the last release. It must be an existing issue specific to Macs.

Could you grab a few details from the Mac where this reproduces? All of it on the page where you see the toast, served from `localhost`.

**1. How many requests does one click send?** DevTools (Cmd-Opt-I) → Network tab → type `enqueue` in the filter box → click Clear → click one `+` icon, once. How many rows appear?

**2. If two rows appear**, click each and from the Headers tab give me the full Request URL for both, the values of `Sec-Fetch-Dest`, `Sec-Fetch-Mode` and `Purpose` (or `X-Purpose`) if present, and what the Initiator column says for each row.

**3. How many click handlers are bound?** In the Console tab, paste this and tell me the number — it should be `1`:

```js
$._data(document, "events").click.filter(h => h.selector === "a.add-song-link").length
```

[...]

PR #878, opened 2026-08-19
fix: queue a song once from the rankings page

safepay:
Found while investigating #877, which turned out to be a Greasemonkey userscript on the reporter's machine rather than a PiKaraoke bug. The double-add below is a separate defect that reproduces with no extension involved.

The rankings page bound its own click handler to `.add-song-link`, on top of the delegated handler `base.html` already binds on the document for every page that renders those links. One click therefore fired both, sending two `GET /queue/enqueue` requests for the same song: the first added it, the second answered `Song is already in the queue: <title>`. Both responses write the same toast element, and the red one arrives last, so that is the message left on screen.

Removed the page's handler and left a note saying where the binding lives. The search page carries the same note already, and the browse page renders only the delegated handler.

Verified in a browser against both versions of the page: one click sends two `/queue/enqueue` requests before this change and one after. The Songs and Add New pages send one per click either way, including across SPA navigation between them.

## Test plan

- [x] Rankings page: click the `+` beside a song not in the queue. One green `<name> added to the queue: <title>` toast, no red one, and one row added to the queue.
- [x] Click the same `+` again. The red `Song is already in the queue: <title>` toast appears, and the queue still holds one row.
- [x] The `Show N` selectors above each chart still reload the page with the chosen row counts, and keep the other selector's value.
- [x] Songs and Add New pages: adding a song still shows the green toast.

### DIFF
--- a/pikaraoke/templates/rankings.html
+++ b/pikaraoke/templates/rankings.html
@@ -14,18 +14,8 @@
 		window.location.href = "{{ url_for('sessions.rankings') }}?" + params;
 	}
 
-	$(function () {
-		// The same endpoint and response shape the browse page uses: the href
-		// already carries the song and ends at "user=", which is resolved at click
-		// time rather than rendered, since it comes from this device's cookie.
-		$(".add-song-link").on("click", function (e) {
-			e.preventDefault();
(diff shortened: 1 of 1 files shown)


safepay:
This is now bundled with #879

PR #879, opened 2026-08-19
fix: queue songs with a POST rather than a GET link

safepay:
Supersedes #878, whose commit is included here as the first of the two.

## Queue a song once from the rankings page

The rankings page bound its own click handler to `.add-song-link`, on top of the delegated handler `base.html` already binds on the document for every page that renders those links. One click fired both, sending two requests for the same song: the first added it, the second answered `Song is already in the queue: <title>`, and since both write the same toast element the red one is what stays on screen.

## Queue with a POST rather than a GET link

The add controls carried the entire request in an `href`, so anything that fetched the link without a click queued the song — a prefetcher, a macOS link preview, a proxy retry, or a browser extension. The click that followed then reported the song as already queued. That is what #877 turned out to be: a Greasemonkey userscript on the reporter's machine, fetching the links it found.

The song now rides in `data-song` and the shared handler posts it to `enqueue_form`, the endpoint the search page has always used. No route or schema changes: both verbs already share `_do_enqueue` and return identical JSON.

The attribute keeps the same percent-encoded path the href carried, and the request body is assembled by hand rather than by jQuery so that value reaches the server byte for byte. That is deliberate — a filename that is not valid UTF-8 cannot survive `decodeURIComponent`, and encoding it afresh would double-encode it. The server therefore sees exactly the string the query parameter used to deliver.

The GET route stays for the documented API.

## Verification

Driven in a real browser against every page that renders these links, counting requests per click:

| Page | Requests per click | Method |
| --- | --- | --- |
| Songs | 1 | POST |
| Rankings (before the first commit) | 2 | GET |
| Rankings (after) | 1 | POST |
| Songs -> Rankings -> Songs by SPA navigation | 1 each | POST |

And the case this is all for: a script that fetches every `a.add-song-link` href on the Songs page now issues **zero** enqueue requests and leaves the queue empty. Previously it queued every song on the page.

### DIFF
--- a/pikaraoke/templates/base.html
+++ b/pikaraoke/templates/base.html
@@ -24,7 +24,9 @@
       cookiePath: "{{ cookie_path }}",
       socketioPath: "{{ socketio_path }}",
       // Empty on the chromeless splash page, which has no singer field.
-      singersUrl: "{% if not blank_page %}{{ url_for('sessions_api.get_singers') }}{% endif %}"
+      singersUrl: "{% if not blank_page %}{{ url_for('sessions_api.get_singers') }}{% endif %}",
+      // Likewise: the splash page renders no add links.
+      enqueueUrl: "{% if not blank_page %}{{ url_for('queue.enqueue_form') }}{% endif %}"
     };
     // Translations for JS (static JS cannot use Jinja2 directly).
     // Gettext output is Markup, so Jinja never autoescapes it. Inside a <script> a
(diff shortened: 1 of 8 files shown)


safepay:
This is the last PR I intend to do for a while. Backlog cleared! Ok to release.

vicwomg:
Ah prefetcher. That makes sense

PR #881, opened 2026-08-20
chore: update translations

safepay:
Final translation gap fill before release. Picks up the singer field text added with KJ mode (#857): `Queue for singer`, `Start a session to queue for singers`, and a reworded `Name of the singer`.

Translated into all 15 locales. No code changes.


vicwomg:
missed the boat. Will follow up in a couple of days in case some things come up from the release being exposed to more people

PR #887, opened 2026-08-20
refactor: group the CLI arguments in --help

safepay:
`pikaraoke --help` was 166 lines: a usage blob listing all 44 flags before describing any of them, then the whole lot under one `options:` heading. Hard to find anything in.

They now sit in seven groups — General, Server and network, Song library and downloads, Splash screen, Background music and video, Playback and audio, Queue — and the usage line is just `pikaraoke [OPTIONS]`.

Nothing about parsing changes. `add_argument_group()` only affects how the help is printed, and I diffed the parsed `Namespace` across nine sample command lines before and after to be sure.

The second commit fixes the `--ytdl-args` help. `--ytdl-args --no-cache-dir` fails with `expected one argument`, because argparse reads the dashes as another PiKaraoke flag — and quoting doesn't help, since the shell strips the quotes first. Only `--ytdl-args="..."` works, so the help now says so.

## Test plan

- [x] `--help` shows a one-line usage and seven headings, every flag listed once
- [x] Flags from different groups still work: `pikaraoke -p 8080 --headless --hide-url --volume 0.5`
- [x] `pikaraoke --ytdl-args="--sleep-requests 1.5" -l 10` downloads a song, with that argument in the logged yt-dlp command

### DIFF
--- a/pikaraoke/lib/args.py
+++ b/pikaraoke/lib/args.py
@@ -161,7 +161,7 @@ def parse_pikaraoke_args() -> argparse.Namespace:
     )
     library.add_argument(
         "--ytdl-args",
-        help="Additional arguments to pass to yt-dlp (as a single string)",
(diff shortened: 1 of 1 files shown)


vicwomg:
looks good, definitely more readable

PR #889, opened 2026-08-21
fix: put the download bar on one honest scale

safepay:
Last part of tidying up the changes that make downloading more robust.

A `+` format selector runs two sequential downloads, and each reported its own 0-100% into a single progress field. The bar filled, snapped back to zero, filled again, then sat at 100% for the whole ffmpeg merge. It read as a crash and a restart, and no restyling fixes it because the number itself was wrong.

Progress now arrives as data through `--progress-template` rather than being scraped from yt-dlp's prose, and `--print before_dl` gives both formats' byte counts before the first byte, so the seam between the video and audio bands is the true size ratio rather than a guess.

The bar only ever draws a percentage yt-dlp reported. The resolve and the merge report none — on a Pi the idle-priority merge runs for tens of seconds — so those animate a travelling highlight in place instead of claiming a number. The fill is a `div` rather than `<progress>` so its width can transition across the 500 ms poll, which is why `<progress>`'s `bulma-dark` rules go with it.

## Test plan

- [X] Queue a download. The bar fills once, left to right, and never snaps back at the video/audio seam.
- [X] Before the first byte and during the merge, a highlight travels in place and the status reads `starting`, then `merging`.
- [X] The bar is white on a dark track, as before.
- [X] Download with high quality off. The bar behaves the same, both settings now fetching separate streams.
- [X] Queue a dead URL. The error card names the error, not a progress line.

### DIFF
--- a/pikaraoke/lib/download_manager.py
+++ b/pikaraoke/lib/download_manager.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import logging
-import re
 import subprocess
 import uuid
 from queue import Queue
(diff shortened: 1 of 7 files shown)


safepay:
Waiting for the heading hierarchy fix.

Issue #894, opened 2026-08-23
[FEATURE] Security step 2 of 2: deny by default


safepay:
**Why:** a guest on the wifi can rename the host's files and drive playback, because a route is open unless someone remembers to protect it. This makes a route closed unless someone marks it open, so nobody has to remember.

Step 1 is #893, which changed how someone becomes an admin. Step 2 changes what happens when they are not one. Step 1 is done and does not block this.

## The problem

Where a route needs protecting, someone writes `if not is_admin()` into it by hand. That works right up until nobody writes anything.

Of 72 routes, 37 have no check. Most are open deliberately — everyone in the room should be able to search and queue from their phone. Some were just missed, and the two look identical in the source. The convention fails open.

What that costs today:

- `POST /batch-song-renamer/rename-song` renames files on disk with no check. Its three neighbours in the same file all have one.
- All seven playback controls answer anyone. The buttons are hidden from guests by a `{% if admin %}` in `home.html`, but `$.get("/skip")` works from any phone on the wifi.
- Two queue error routes are ungated, sitting next to gated siblings.

## The fix

Refuse everything by default, and mark the routes the room is meant to reach with `@public`, enforced by a `before_request` gate in a new `lib/auth.py`.

A route added without a decision is then closed rather than open. The decision sits on the route instead of in a list of names in another file, and it cannot go stale when a view gets renamed.

## How it lands

1. **Machinery.** Add the gate, mark all 37 currently-open routes `@public`, change nothing else. Provably a no-op, so it reviews quickly.
1. **Audit.** Every marker from part 1 is a claim; delete the false ones. At least ten are false — the playback controls, the renamer, the two queue routes. Check each route's UI, not just its body: that is exactly how the playback gap survived.
1. **Cleanup.** Delete the 18 inline checks and the two `before_request` allowlists. Leave the five `admin=is_admin()` template arguments alone — those decide whether a button renders, and are not authorization.

## The test that holds it together

```python
def test_every_endpoint_is_decided():
    """Every route is public by explicit mark, or host-only. No third state."""
    public = {
        rule.endpoint
        for rule in app.url_map.iter_rules()
        if getattr(app.view_functions[rule.endpoint], "pika_public", False)
    }
    assert public == EXPECTED_PUBLIC_ENDPOINTS
```

A golden list in the test file. Opening a route to the room then means editing it, which is a line in a diff someone has to agree with.

## Notes

[...]

Issue #893, opened 2026-08-23
[FEATURE] Security step 1 of 2: set the admin password without restarting


safepay:
**Why:** the host cannot change the admin password without stopping PiKaraoke and relaunching it with a different command line. This makes it settable from the Info page while it is running.

Two pieces of work, split across two issues.

**Step 1**, this one, is about how someone becomes an admin. **Step 2** is about what an admin is allowed to do, and what a guest is not. They are separate because they break in different ways and can land independently. Step 2 is #894.

## Who this protects against

One guest on the party wifi, with the QR code and a browser. What is at risk is the host's files, whether the box stays running, and the play history.

If no admin password is set, everyone is an admin. That is the right default for a karaoke box on the living room TV, and neither step changes it. All of this only matters once a host decides to set a password.

## The problem

You could only set the admin password with `--admin-password` when starting PiKaraoke. Changing it, or clearing it, meant quitting and relaunching with a different command line. In practice that means editing whatever starts PiKaraoke on that machine — a shortcut, a service file, an autostart entry — which is not something a host should have to do to change a password.

Two smaller problems came with it. The `admin` cookie contained the password itself, in plain text, valid for 90 days, so it sat in every browser that had ever logged in and went over the network on every request. And the session signing key was regenerated at every launch, which quietly threw away each visitor's language choice on restart.

## What step 1 does

Branch: [safepay/pikaraoke@fix/admin-password-without-restart](https://github.com/safepay/pikaraoke/tree/fix/admin-password-without-restart). No PR yet due to a large number or recent PR's!!

- You can set, change or clear the password from the Info page while PiKaraoke is running. No restart.
- The password is stored as a salted PBKDF2 hash in a `[SECRETS]` section of `config.ini`, so it is never written down in readable form.
- Logging in gives the browser a signed session instead of the password, so the password itself never leaves the server.
- Changing the password logs every other device out. The device you changed it on stays logged in.
- The signing key is saved rather than regenerated, so logins and language choices survive a restart.
- `--admin-password` still works, and now sticks. That is how you reset a password nobody remembers. Pass the flag with no value to clear it.

`is_admin()` behaves exactly as it did before, which is what keeps step 2 independent of this work.

## Why this went first

Today `is_admin()` is called by hand in 18 places. After step 2 it runs on every request instead.

So a bug in it right now costs the host their admin panel. After step 2, the same bug would stop every guest searching and queueing. It was cheaper to rework it while the consequences of getting it wrong were still small.

[...]

PR #895, opened 2026-08-23
fix: require POST for the playback control routes

safepay:
The seven playback controls change state on `GET` — `/skip`, `/pause`, `/restart`, `/transpose/<n>`, `/volume/<n>`, `/vol_up` and `/vol_down`. They are the last state-changing `GET`s after #886.

This precedes the work in #893 and #894.

On its own this buys nothing, since they are still ungated and there is no privilege to steal yet. It is the safe half of the order: briefly POST-only and ungated is no worse than today, whereas gating first would leave them privileged and reachable by a link.

Every caller was already AJAX, so no markup moves — seven `$.get` become `$.post` in `home.html` and one in `queue.html`.

Also deletes the `.playback-btn` handler in `queue.html`, which binds to a class no markup carries, like the dead handlers #886 removed.

Also drops `GET /enqueue` and its `EnqueueQuery` schema, missed by #886 because it sits in `routes/queue.py`. Deleted rather than converted: nothing calls it, and a POST route no caller posts to is still dead code.

`tests/unit/test_state_changing_methods.py` goes from eleven endpoints to nineteen.

## Test plan

Run with `--admin-password` set and log in as admin.

- [x] Home page, Player Control: restart, play/pause and skip all work, and the restart and skip confirmations still appear and still do nothing when cancelled
- [x] Home page: the volume slider changes the volume, and the up and down buttons work
- [x] Home page: transpose confirms, applies, and resets the slider to zero
- [x] Queue page: open the now-playing options modal and check restart, pause and skip
- [x] `http://<host>:5555/skip` typed into the address bar returns 405 instead of skipping the song

### DIFF
--- a/pikaraoke/routes/controller.py
+++ b/pikaraoke/routes/controller.py
@@ -12,7 +12,7 @@ _ = flask_babel.gettext
 controller_bp = Blueprint("controller", __name__)
 
 
-@controller_bp.route("/skip")
+@controller_bp.route("/skip", methods=["POST"])
 def skip():
     """Skip the currently playing song."""
     k = get_karaoke_instance()
@@ -21,7 +21,7 @@ def skip():
     return redirect(url_for("home.home"))
 
 
-@controller_bp.route("/pause")
+@controller_bp.route("/pause", methods=["POST"])
 def pause():
     """Toggle pause/resume playback."""
     k = get_karaoke_instance()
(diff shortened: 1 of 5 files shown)


vicwomg:
@safepay haha was going to say. "Feel free to merge" things because I'm not sure of the order sequence. I think I don't need to be as involved in the review process in general these days.

safepay:
Yes. I know what you mean! But I see your independent input as critical, particularly on UI changes.

However, if there are smaller PR's that tidy up code, let me know if I can merge them without your review. I will be very careful.

safepay:
And a second point - I also explicitly understand the need to let a large release "sit' for a while without anything other than bugfix commits. So I would not commit anything that impacted this.

safepay:
@vicwomg So I would merge 889 and 896, then do a patch release because that effectively makes PiKaraoke pretty seamless for an install. I.e. There are no CGI flags beyond --headless that are required to operate the program.

All future PR's will be larger and require a separate release.

PR #900, opened 2026-08-26
feat: refuse every route not marked public

### LINKED ISSUE #894: [FEATURE] Security step 2 of 2: deny by default

safepay:
**Why:** a guest on the party wifi can rename the host's files and drive playback from their phone, because a route is open unless someone remembers to protect it. After this, a route is closed unless someone marks it open, so nobody has to remember.

Implements #894, the second of the two security steps, on top of #896.

## What changed

- **New `lib/auth.py`.** `@public` marks an endpoint the room may reach, `@answers_json` marks one whose refusal must be JSON, and `install_auth_gate(app)` refuses everything else before the view runs. Two bare markers, neither taking arguments.
- **33 endpoints marked `@public`**, verified against the running `url_map` rather than by reading route files — that is how the playback gap survived the last audit.
- **Eight routes closed:** the seven playback controls, whose intent lived only in a `{% if admin %}` in `home.html`, and `POST /batch-song-renamer/rename-song`, which renamed files on disk with no check at all while its three siblings each had one.
- **Two more closed by auditing the markers themselves:** `/metadata/auto-format` and `/metadata/suggest-names` are called only from the rename page, which is host-only. `suggest-names` makes an outbound iTunes request per call, so any phone on the wifi could have the box make one on demand.
- **A guest opening `/info` no longer fetches the host's stats.** Two `$.getJSON` calls sat outside the admin block, so every guest paid two refused round trips for a panel that does not render for them.
- **20 inline `is_admin()` checks and both blueprint allowlists deleted.** The gate runs ahead of the view, so leaving them would leave 20 blocks of dead code behind it.
- **Five refusal shapes collapse to one:** JSON 403 under `/api`, for XHR, and where `@answers_json` declares it; flash and redirect home otherwise.
- **The gate checks the marker before `is_admin()`.** Reading the password hash goes through `PreferenceManager.get`, which reparses `config.ini` on every call — 0.4ms on a desktop, more on a Pi. Ordering the cheap checks first keeps that off the 33 public endpoints, including every HLS segment served during playback.
- **The twelve bespoke refusal strings are gone**, replaced by one generic message. Every one of the thirteen routes that carried one hides its button behind `{% if admin %}` — including `/sessions`, whose nav link turned out to be gated in `base.html` too — so the wording only ever reached someone hand-crafting a request, or a host whose session had expired, and the generic line serves both as well. Dropping them buys back twelve msgids across 30 locales, a `_lazy` alias in five route files, and a `-k _lazy` extractor flag that existed solely to keep `lazy_gettext` strings in `messages.pot`.
- **`@answers_json` rather than sniffing.** `/library_stats`, `/sync_library`, `/info/stats`, `/queue/reorder`, the two `/metadata` lookups and `rename-song` answer JSON but do not sit under `/api`. Without the marker a refused XHR gets a 302 that jQuery follows, handing the caller a 200 and a page of home-page HTML — a wrong answer, not just an ugly one. Guessing from `X-Requested-With` instead works today only because every caller happens to be jQuery. Mounting these under `/api` retires the marker; that is the routes/API split plan, not this PR.
- **Golden list in `tests/unit/test_auth_gate.py`.** Opening a route to the room now means editing a list, which is a diff a reviewer has to agree with.
- **Route test apps install the gate.** `make_route_app` takes an `admin` flag backed by a `StubAdminAuth`; authorization is mocked at `pikaraoke.lib.auth.is_admin`, while a route module's own `is_admin` still only decides what its template renders.

## Deliberate behaviour changes beyond the gate

- `queue_edit`'s bare `"Unauthorized"` flash and the batch renamer's silent redirect to `/browse` are gone; both now get the generic refusal, as do the eleven routes that had bespoke wording.
- A guest who posts a rename lands on home rather than back at the referrer.
- The four `admin=is_admin()` template arguments are untouched — they decide whether a Delete button renders, not who may call the route.
- `POST /logout` stays public although its button is admin-only. It clears only the caller's own session, so refusing a guest buys nothing, and gating the action that ends a session on having a valid one is circular. The reason is recorded at the marker.

`DELETE /queue/downloads/errors/<id>` and its retry sibling stay public. The plan listed them as forgotten gates; the Retry and Dismiss buttons render outside any `{% if admin %}`, so the download-error list is genuinely offered to the room.

## Test plan

- [x] With an admin password set and no admin session: search, queue a song, watch it play, open `/history` and `/rankings`
- [x] The Player Control box is absent, and `POST /skip` from the console refuses rather than skipping
- [x] `/files/edit`, `/sessions` and `/batch-song-renamer` all bounce to home with a message
- [x] As admin: skip, pause, volume, transpose, rename a song, clear the queue, change a preference, sync the library
- [x] The Info page's system and library stats still load, and Update yt-dlp still starts
- [x] Dismiss and Retry on a failed download still work from a guest's phone
- [x] On the rename page as admin, Auto-format and the name suggestions both still return results
- [x] As a guest, `/info` loads with no console errors and no refused requests in the network tab
- [x] With the interface set to another language, a refused action shows the refusal — in English until the catalogs are regenerated, since the generic line is the one new string


safepay:
@vicwomg Even though this is ready to merge, I would recommend doing a bug fix release FIRST. Then we can do more refactoring/feature work.

vicwomg:
> @vicwomg Even though this is ready to merge, I would recommend doing a bug fix release FIRST. Then we can do more refactoring/feature work.

Ok, let me know when a good time to do that would be.

safepay:
Ready to release now. All translations are updated.

safepay:
@vicwomg Don't release yet! There is now a bug with the YT previews since YT turned off the pre-merged formats. I think I'll have to remove that option and just open the videos in the browser/app.

PR #903, opened 2026-08-27
feat: accept a directory of background videos

### LINKED ISSUE #533: [FEATURE] Randomize Splash Screen Background
**Is your feature request related to a problem? Please describe.**
N/A

**Describe the solution you'd like**
A feature where you can specify a custom directory for the splash screen background video.
If a folder is specified, it will pick a random background video from the folder when displaying the splash screen.
If you specify a specific file, you could leave behavior as it is now where just that specific background video will play.

**Describe alternatives you've considered**
N/A - Could make multiple bats but it requires restarting pikaraoke each time

**Additional context**
N/A

### LINKED ISSUE #544: [FEATURE] Playlist of music clips playing on the background when queue is empty
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
On the info page, add the possibility to put on a youtube playlist url to be randomly played on the background video when the queue is empty. This way, the admin can set a playlist with music clips to be played while there's no one singing.

Maybe the system can "rotates" the download/play/delete songs, using the ytdlp with a temporary folder to download a video from the playlist and while the video is being played it downloads another one and when it finishes playing, the system deletes the song and so on.

Sorry @vicwomg for I'm having no time to implement this. If no one attacks this, maybe I can give it a try in the future.

safepay:
**Why:** A host with a folder of backgrounds can point `--bg-video-path` at it and get a different one each time the room goes quiet, instead of naming one file and editing the command line to change it. Closes #533.

## Changes

- `--bg-video-path` takes a directory: one of its `.mp4`, `.m4v` and `.webm` videos is picked at random when the splash screen returns, and loops until the next song.
- One video per idle spell, never a swap partway. Backgrounds cut to tile are often ten seconds long, so playing a folder end to end flicked between them.
- The candidates are settled at render, each with its own URL: the route answers range requests, so picking per request would serve one video's bytes against another's length. Shuffled before the fifty-video cap, so a larger folder still offers everything.
- `/stream/bg_video` becomes `/stream/bg_video/<file>`, served by `send_from_directory` now the name comes from the URL. `..\`, `../` and a drive letter are refused, as for `/bg_music/<file>` after #884.
- Naming one video serves that video alone, not its folder. `--dolphly` unchanged.
- The content type is no longer hardcoded to `video/mp4`, so a `.webm` is announced as one. The dead `<source>` in `splash.html` goes too.
- `Background video not found ... Setting to None` now drops the path, which it never did.
- The page gets the video list in place of `has_bg_video`, a flag `Karaoke`'s `night_sea.mp4` fallback made permanently true.

## Test plan

- [x] A folder of three videos: empty the queue several times, and more than one of the three appears.
- [x] One idle spell loops one video for several minutes, with no swap and no seam on a ten-second background.
- [ ] A `.webm` plays on Safari or iOS.
- [x] An empty directory renders the splash with no video; a missing path prints the message and falls back to the default.
- [x] A single file, and `--dolphly`.


safepay:
Please run the Safari/iOS test.

PR #917, opened 2026-08-30
refactor!: move every JSON route under /api

### LINKED ISSUE #923: [REFACTOR] Decide a route's medium by its path
**Why:** whoever maintains this can answer "what does an unauthorized request get?" by reading the route's path, and a guest's blocked action fails visibly instead of appearing to work.

## Problem

- `_refuse` weighs four signals to choose between a JSON 403 and a redirect: the path, `X-Requested-With`, `Accept`, and a hand-applied `@answers_json` marker.
- It gets it wrong silently. With an admin password set, an unauthorized `POST /skip` answers `200 OK` with home-page HTML, which every fire-and-forget `$.post` reads as success.
- Nothing tells the next route's author which signal to set, and nothing catches them setting none.
- The OpenAPI spec declares no security scheme, so `/apidocs` asserts by omission that the whole API is open.

## Solution

- One rule replaces the four: everything answering JSON sits under `/api`, and no page does. `_refuse` reads the path and nothing else.
- Twenty-nine routes gain `/api`; `@answers_json`, the `Accept` fallback and the `X-Requested-With` sniff go.
- A test holds the tree to the rule both ways off the live `url_map`.
- `document_auth` requires the cookie where the gate already does, so the spec names the credential instead of hiding it.

Breaking: those twenty-nine move without an alias. An alias is not available, as the gate reads the path to pick its refusal.

## Done when

- [ ] An unauthorized JSON route gets a JSON 403, never HTML
- [ ] CI fails if a JSON route is added outside `/api`, or a page inside it
- [ ] `/apidocs` says which operations need a login and how to get one
- [ ] The `Accept`, `X-Requested-With` and `@answers_json` signals are gone

safepay:
**Why:** Today a refused `POST /skip` answers `200 OK` with the home page, so no caller — the queue page's own buttons, or anything scripted against PiKaraoke — can tell a denial from a success. The gate chose that answer from four competing signals. It now reads one, the path: every JSON route sits under `/api` and refuses in JSON, and `/apidocs` drops from 57 paths to 34, each declaring whether it needs a login.

## The rule

A route's medium is its path: everything answering JSON sits under `/api`, and no page does.

- **Twenty-nine routes gain `/api`**, in the modules they already live in. `url_for` follows for free; eight hardcoded paths needed editing.
- **`_refuse` keeps only the path check.** The `Accept` fallback, the `X-Requested-With` sniff and `@answers_json` all go, and so does the header two callers were still sending for them.
- **`test_route_mediums.py` holds the tree to it** both ways off the live `url_map`, following bare-name calls into the view's own module — without that step `enqueue_form` is invisible, its `jsonify` living in `_do_enqueue`.
- **The blueprint lists move to `routes/__init__.py`**, so `app.py` and the tests cannot disagree.

## What it took out

- **Eight endpoints stop claiming `text/html`**: `json.dumps` becomes `jsonify`, so five callers stop hand-parsing the string. One of them was already parsing a parsed object.
- **`queue/edit` sheds an unreachable HTML fallback**: POST-only with jQuery callers made `is_ajax` always true, stranding a redirect, a flash and ten translated labels behind a branch nothing could take.
- **The playback verbs, `addrandom` and `clear_preferences` stop redirecting.** Each answered a click with a 302 its caller discarded. Two had flashes consumed into that discarded response, so "Ran out of songs!" had never once reached anyone; it does now.
- **Dead weight**: the `werkzeug.__version__` shim across eleven test files, two SPA exclusions matching no route.

## Modules

- **`library_api.py` splits out of `admin`**: `/api/library_stats` and `/api/sync_library` sat there because the info page calls them, not because they belong beside `quit` and `auth`. URLs unchanged.
- **`auth_api.py` gives a program somewhere to log in**: `/auth` is a browser form that answers 302 whether the password was right or wrong. The JSON twin answers 200 or 401. The session grant behind both moves to `lib/auth.py`, which held three copies of it.
- **`background_music` goes internal**: `bg_playlist` is polled by `splash.js` and nothing else, the same readership as `splash.score_phrases` beside it.

## /apidocs

Falls out of the rule rather than driving it. 34 paths where there were 57, all but `/queue` and `/search` under `/api`, each saying whether it needs a login and how to get one.

- **The spec names the credential it was hiding.** Nothing declared a security scheme, so it asserted by omission that the whole API was open. `@public` now clears an operation's `security` as well as opening the gate, and `document_auth` requires the cookie on the other 24 of 37 operations, off the marker the gate already reads.
- **`TestSpec` holds the spec to the gate**, walking every documented operation. The two are read at different times — the gate off the live view, the spec at registration — so a `@public` above its `@route` opens the gate while the spec still claims a login. Nothing else catches that.

`/queue` and `/search` are the last two pages listed. Both need a module split, which is `queue_api.py` and `search_api.py` and its own branch.

## Breaking

Twenty-nine endpoints moved under `/api` with no alias — `/now_playing`, `/get_queue`, `/enqueue`, `/queue/*` and the playback verbs are the ones anyone is likely to have scripted against. An alias is not available as a softener: the gate reads the path to choose its refusal, so a legacy `/skip` would be a JSON route outside `/api`, and `test_route_mediums.py` fails on it by design. `queue/edit` drops its `message` field, the playback verbs, `addrandom` and `clear_preferences` answer JSON rather than a redirect, and eight endpoints send `application/json` where they sent `text/html`.

The stream and image URLs are unchanged and still work — they have only left `/apidocs`.

## Test plan

Eight paths are hand-built in templates and JS, where `url_for` cannot fail loudly at render. These exercise them.

- [X] Splash: now playing updates, background music shuffles
- [X] `/files` song edit: "suggest names" and "auto format" both answer
- [X] Add a song from browse and from history; both notifications name it
- [X] Queue: reorder a row, dismiss a download error, then cog menu move/delete and "Clear all"
- [X] Queue: "Add 3 random songs" adds them; with every song already queued it says "Ran out of songs!"
- [X] `/info`: song count loads, "Sync library" scans, and "Clear preferences" confirms, says so, and brings the page back on defaults with every toggle still live
- [X] Home controls — skip, pause, volume, transpose — and the queue's now-playing modal: restart, pause, skip
- [X] Logged out with a password set: `/api/skip` and `/api/library_stats` are JSON 403s, `/quit` redirects
- [X] `--enable-swagger`: `/apidocs` lists 34 paths, and `POST /api/skip` shows a padlock where `POST /api/enqueue` does not
- [X] With a password set: `curl -i -X POST .../api/auth -H 'Content-Type: application/json' -d '{"admin_password":"..."}'` returns 200 and a `Set-Cookie`, a wrong one returns 401, and the cookie then opens `/api/skip`
- [X] `--base-path /karaoke`: splash and queue still poll


safepay:
Ready to merge.

safepay:
The frikkin' /apidocs got me again!

safepay:
100% finished and ready to merge.

Issue #899, opened 2026-08-26
[BUG] Downloaded song doesn't have audio stream


e88z4:
**Describe the bug**
PLEASE FILL THIS FORM OUT ENTIRELY for a response. 

I recently start to have issue downloading all songs from youtube. The song download is successful but when I queue the song, the ffmpeg stream didn't play at all (no audio, no video)

**To Reproduce**
1. Add new song 
2. Queue the song
3. Play the song

Error message:
[2026-08-25 20:37:32] INFO: ARM platform detected, using h264_v4l2m2m hardware encoder
[2026-08-25 20:37:32] ERROR: FFmpeg exited with code 234
[2026-08-25 20:37:32] ERROR: Failed to prepare stream
[2026-08-25 20:37:32] ERROR: Failed to prepare stream

**Expected behavior**
Audio and video stream are being played.

**Platform (please complete the following information):**
 - Hardware: Raspberry pi 4
 - OS: Debian Trixie
 - OS version: 13.6
 - Browser chromium
 - Version 1.22.0

**Additional context**
I investigated the downloaded file, the song downloaded has only video stream, no audio stream at all. I was able to confirmed by playing the audio manually in VLC. 
Is this a known issue? I have over 300+ song downloaded, this bug started the last 5 songs that I downloaded recently within the last 7 days.

e88z4:
This is a sample of log of a downloaded song using youtubedl version 2026.08.19. I verified using VLC that the downloaded mp4 has video stream, no audio.

```
[2026-08-25 21:53:45] DEBUG: yt-dlp command: /home/felix/.local/share/uv/tools/pikaraoke/bin/python3 -m yt_dlp -f mp4 -o /home/felix/pikaraoke-songs/%(title)s---%(id)s.%(ext)s -S vcodec:h264 --compat-options filename-sanitization https://www.youtube.com/watch?v=eXqBhDAlVCc
[2026-08-25 21:53:51] DEBUG: yt-dlp output: WARNING: "-f mp4" selects the best pre-merged mp4 format which is often not what's intended.
         Pre-merged mp4 formats are not available from all sites, or may only be available in lower quality.
         To prioritize the best h264 video and aac audio in an mp4 container, use "-t mp4" instead.
         If you know what you are doing and want a pre-merged mp4 format, use "-f b[ext=mp4]" instead to suppress this warning
[youtube] Extracting URL: https://www.youtube.com/watch?v=eXqBhDAlVCc
[...]

PR #943, opened 2026-09-08
fix: require an audio track in the download fallbacks

GeekShopAdmin:
**Why:** A video whose only audio is opus-in-webm downloads video-only, exits 0, enters the library, then fails at playback with `Skipping track`. Follow-up to #899, which fixed `-f mp4` but left the `best[...]` fallbacks still admitting a video-only format.

## Changes

- Add an any-audio pair (`+bestaudio`) before falling back to a single format, for videos whose only audio is webm.
- Require `acodec!=none` in both fallbacks, so a video-only format can no longer satisfy them.
- Keep `bestaudio[ext!=webm]` first, so audio is still stream-copied where possible.

A video offering only webm-muxed audio now fails visibly instead of downloading silently. The `.mkv` the new pair produces needs no container forcing — it is already in `song_list.VALID_EXTENSIONS`, and `lib/ffmpeg.py` already transcodes it.

## Test plan

- [x] Download `2014hFU6I9s`, whose only audio is opus-in-webm — yt-dlp merges an avc1 video with the webm audio, `ffprobe` shows `video,audio`, and it plays. Before: format 136, video-only, `ffprobe` showed `video`.
- [x] Download a video with m4a available — unchanged, still stream-copies.
- [x] Queue and play both — no transcode error, no `Skipping track`.
- [x] Running on a Pi 4 all day via `--ytdl-args`, no side effects seen.

Existing silent files can't be repaired; this finds them:

```bash
find /path/to/songs -type f \( -iname '*.mp4' -o -iname '*.mkv' -o -iname '*.webm' \) -print0 |
while IFS= read -r -d '' f; do
  [ -z "$(ffprobe -v error -select_streams a -show_entries stream=index -of csv=p=0 "$f")" ] && echo "$f"
done
```

### DIFF
--- a/pikaraoke/lib/youtube_dl.py
+++ b/pikaraoke/lib/youtube_dl.py
@@ -174,10 +174,17 @@ def build_ytdl_download_command(
     height = 1080 if high_quality else 720
     # The capped fallback matters because -S sorts on resolution: without it, a video with
     # no avc1 DASH pair lands on a 1080p HLS variant no matter what the cap says.
+    # bestaudio[ext!=webm] stays first so the audio can be stream-copied, but YouTube
+    # publishes opus-in-webm as the only audio for some videos: that pair then matches
+    # nothing, so pair with any audio before falling back to a single format. Every
+    # fallback requires acodec!=none -- a video-only format otherwise satisfies
+    # best[...], downloads silently, exits 0, and only fails later at -map 0:a, which
+    # is a successful download that will not play.
     file_quality = (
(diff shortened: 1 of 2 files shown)


GeekShopAdmin:
Comment down to three lines, PR body cut to Why/Changes/Test plan. Selector itself unchanged.

Also pushed a test fix: `test_pre_merged_fallback_honours_the_cap_before_giving_up_on_it` still asserted the old tiers, so this branch was failing two tests. Its checks never showed that — the workflow is awaiting maintainer approval. 1404 tests and black pass now.

Issue #519, opened 2025-07-25
[BUG] proxy error "not an accepted origin"


fuentes81081:
I'm running the pikaraoke docker image on a linux server.  I have an nginx reverse proxy setup pointing to the internal IP.  Everything works fine when using the browser on the internal IP.  However, when accessing pikaraoke via the proxy, the pikaraoke logs state "https://my.domain.org:1234/ is not an accepted origin."

When using the browser inspector, it revealed many code 400 errors when trying to reach https://my.domain.org:1234/socket.io

**To Reproduce**
1. Go to the proxy address ie. https://my.domain.org:1234/splash and add a song to the queue
2. Videos will not play.
3. Browser inspector shows repeated HTTP 400 errors to https://my.domain.org:1234/socket.io
4. pikaraoke logs reveal "https://my.domain.org:1234/ is not an accepted origin."

**Expected behavior**
Videos should play when added to the queue.

**Platform:**
 - Hardware: Linux server
 - OS: Ubuntu 64-bit
 - OS version: 24.04 
 - Browser: Chrome
 - Using docker image vicwomg/pikaraoke:latest

**Additional context**
I found the socketio documentation here: https://flask-socketio.readthedocs.io/en/latest/deployment.html#cross-origin-controls

>If an incoming HTTP or WebSocket request includes the Origin header, this header must match the scheme and host of the connection URL. In case of a mismatch, a 400 status code response is returned and the connection is rejected.

I was able to resolve these errors by changing line 48 in [app.py](https://github.com/vicwomg/pikaraoke/blob/909f9f4808439278d2793d1f4f3c2a8217807286/pikaraoke/app.py#L48) to
`socketio = SocketIO(async_mode="gevent", cors_allowed_origins="*")`

I understand this is not the safest solution.  I propose setting the value based on a docker environment variable, like:

`docker run -e ALLOWED_ORIGINS=https://my.domain.org:1234`

vicwomg:
Have you tried the approach outlined here?
https://github.com/vicwomg/pikaraoke/issues/520

fuentes81081:
I haven't done that exact approach because I'm using nginx, not caddy.

I think the caddyfile is analogous to the site conf in nginx, here's what mine looks like

```
server {
    listen 1234 ssl default_server;
    listen [::]:1234 ssl default_server;
    server_name my.domain.org;
    include /config/nginx/ssl.conf;

    location / {
        proxy_pass http://192.168.68.51:5555;
[...]

PR #944, opened 2026-09-08
fix: keep loopback an accepted socket.io origin when --url is set

GeekShopAdmin:
**Why:** Setting `--url` makes that URL the only accepted socket.io origin, so a headless host running its own kiosk browser on loopback has its socket refused — the splash sits on the logo indefinitely and the only clue is `<origin> is not an accepted origin`. #519 fixed the reverse-proxy direction by making `--url` the allowed origin; both origins need to be accepted, not one or the other.

## Changes

- Add the loopback origins (`localhost`, `127.0.0.1`, `[::1]`) alongside `--url`.
- Normalise `--url` through `urlsplit` to `scheme://host[:port]`. `karaoke.py` supports a base path in `--url`, and an `Origin` header never carries one, so `--url https://host/karaoke` could never match anything a browser sends — broken today, independently of this PR.
- Keep `cors_allowed_origins = None` when `--url` is unset, preserving engineio's same-origin default. Always passing a list would narrow that default from "same origin" to "only these" and break LAN access for anyone who never sets `--url`.

Does not cover the LAN address or `--prefer-hostname` while `--url` names a public host — that needs the `--allowed-origins` flag asked for in #519. Happy to do that separately.

## Test plan

- [x] `--headless -u https://<host>` with kiosk Chromium on `http://localhost:5555/splash` — splash connects and plays. Previously stuck on the logo all day.
- [x] Same box reached over the public URL — still works.
- [x] No `--url` — LAN access unchanged.
- [ ] `--url https://host/karaoke` with a base path — not exercised on hardware.

### DIFF
--- a/pikaraoke/app.py
+++ b/pikaraoke/app.py
@@ -13,7 +13,7 @@ import logging
 import os
 import sys
 from pathlib import Path
-from urllib.parse import quote
+from urllib.parse import quote, urlsplit
 
 import flask_babel
 from flask import Flask, request, session
@@ -58,20 +58,18 @@ from gevent.pywsgi import WSGIServer
 
 args = parse_pikaraoke_args()
 socketio_path = f"{args.base_path}/socket.io" if args.base_path else "/socket.io"
-# --url replaces engineio's same-origin default rather than adding to it, so setting it
-# makes that URL the ONLY accepted origin. A headless host running its own kiosk browser
-# reaches the splash on loopback, and that connection is then refused: the splash stays
-# on the logo with no play events and nothing is logged but "is not an accepted origin".
-# Keep None when --url is unset, so the same-origin default is preserved for everyone
-# who is not overriding the URL.
+# Setting cors_allowed_origins REPLACES engineio's same-origin default, so --url alone
+# locks out the kiosk browser on loopback ("is not an accepted origin"). Origin headers
+# never carry a path, hence the normalise.
+cors_allowed_origins = None
 if args.url:
+    split_url = urlsplit(args.url)
     cors_allowed_origins = [
-        args.url,
+        f"{split_url.scheme}://{split_url.netloc}",
         f"http://localhost:{args.port}",
         f"http://127.0.0.1:{args.port}",
+        f"http://[::1]:{args.port}",
     ]
-else:
-    cors_allowed_origins = None
 socketio = SocketIO(
     async_mode="gevent", cors_allowed_origins=cors_allowed_origins, path=socketio_path
 )
--- a/pikaraoke/app.py
+++ b/pikaraoke/app.py
@@ -58,7 +58,23 @@ from gevent.pywsgi import WSGIServer
 
 args = parse_pikaraoke_args()
 socketio_path = f"{args.base_path}/socket.io" if args.base_path else "/socket.io"
-socketio = SocketIO(async_mode="gevent", cors_allowed_origins=args.url, path=socketio_path)
+# --url replaces engineio's same-origin default rather than adding to it, so setting it
+# makes that URL the ONLY accepted origin. A headless host running its own kiosk browser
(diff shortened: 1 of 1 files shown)


safepay:
Thanks - this is the right minimal fix for the stuck kiosk, and keeping `None` when `--url` is unset is the important half.

Two things before it goes in.

**The LAN case isn't blocked for the reason given.** `get_ip()` in `lib/network.py` is a free function taking a platform string, so it is callable at import. The real awkwardness is different: on a Pi, `karaoke.py` deliberately avoids it and loops on `hostname -I` for up to 30s because the address may not exist yet at startup, so an import-time call could freeze the `127.0.0.1` fallback. Worth staying minimal, but for that reason rather than availability.

**`--url` can carry a path, and an `Origin` header never does.** `karaoke.py` branches on `split_url.path` for base-path deployments, so `--url https://host/karaoke` is a supported shape - and as an allowed origin it is a string no browser can ever match. Anyone on a reverse proxy with a base path is broken today, before this PR. Since you're already touching the line, normalising through `urlsplit` to `scheme://host[:port]` would fix it in passing.

On the `--allowed-origins` question: worth hearing what @vicwomg thinks, but without it the LAN-address and `--prefer-hostname` cases (origin becomes the FQDN) stay broken, so this closes the kiosk case rather than the class.

Minor: `http://[::1]:<port>` isn't covered, and the comment could lose a couple of sentences - the measured symptom and the "replaces rather than adds" quirk are the parts the code can't say.

GeekShopAdmin:
Both corrections taken, and the second one is a bug I'd have shipped.

**`get_ip()`.** You're right and my reason was wrong — `get_ip(platform)` is a free function and callable at import; I said otherwise without checking. The reason you give is the real one and a better one: `karaoke.py` loops on `hostname -I` for up to 30s precisely because the address may not be up yet, and `get_ip` returns `127.0.0.1` when detection fails, so an import-time call risks baking that fallback into the origin list. Staying minimal for that reason instead.

**`--url` with a path.** This is the one worth fixing here. `karaoke.py:444` branches on `split_url.path not in ("", "/")`, so `--url https://host/karaoke` is a supported shape — and an `Origin` header never carries a path, so that string can never match anything a browser sends. As you say, anyone on a reverse proxy with a base path is broken today, independently of this PR. Normalising while I'm on the line:

```python
if args.url:
    split_url = urlsplit(args.url)
    cors_allowed_origins = [
        f"{split_url.scheme}://{split_url.netloc}",
        f"http://localhost:{args.port}",
        f"http://127.0.0.1:{args.port}",
        f"http://[::1]:{args.port}",
    ]
else:
    cors_allowed_origins = None
```

`[::1]` added — also your catch. `app.py` currently imports only `quote` from `urllib.parse`, so that import line grows too.

And the comment down to the parts the code can't say for itself:

```python
# Setting cors_allowed_origins REPLACES engineio's same-origin default rather than
# adding to it, so --url alone makes that URL the only accepted origin and a kiosk
# browser on loopback is refused -- the splash sits on the logo and the only clue is
# "is not an accepted origin". Origin headers carry no path, hence the normalise.
```

On `--allowed-origins`: agreed that's @vicwomg's call, and agreed on the scope — this closes the kiosk case, not the class. The LAN-address and `--prefer-hostname` cases (where the origin becomes the FQDN) still need it. Happy to do that as a follow-up if he wants it.

GeekShopAdmin:
Both taken, and the second was a bug I'd have shipped. Pushed: `urlsplit` normalise to `scheme://netloc`, `[::1]` added, `else` branch dropped, comment down to three lines. PR body cut to Why/Changes/Test plan. 1404 tests and black pass.

You were right about `get_ip()` too — it's a free function and callable at import; I'd said otherwise without checking. Staying minimal for the reason you gave instead: the `hostname -I` loop exists because the address may not be up yet, and `get_ip` falls back to `127.0.0.1`.

Issue #942, opened 2026-09-07
[BUG] Songs stall ~1s before the end and the queue never advances on Chromium 150+ (CDG video stream is shorter than the audio)


GeekShopAdmin:
**Describe the bug**

Playback stops about a second before the end of every song and the `ended` event never fires. `endSong()` is therefore never called, no `end_song` reaches the server, and the queue holds indefinitely. The first occurrence on my machine sat there for roughly 40 minutes until someone pressed skip.

Two separate defects combine to produce it:

1. **The CDG transcode leaves video shorter than audio.** CDG graphics routinely stop seconds before the music does (lyrics finish, instrumental outro plays on). `build_ffmpeg_cmd` maps both streams with no padding and no `-shortest`, so the output has audio running past the end of video. Measured gaps in my library: 7 to 17 seconds.
2. **From Chromium 150, the browser stops playback at the shorter track** instead of playing through the mismatch. Chromium 143 played the same files to completion. This is what turns defect 1 into a wedged queue.

Because it is the mismatch that matters and not the packaging, this is not CDG-only. YouTube-downloaded MP4s have their streams copied rather than re-encoded, carry a mismatch of roughly a second, and stall about a second short.

**To Reproduce**

1. Run PiKaraoke on a machine whose Chromium is 150 or newer.
2. Queue any MP3+CDG song.
3. Let it play to the end.
4. Playback freezes a second or more before the end, `ended` never fires, and the queue does not advance.

**Expected behavior**

The song plays to its end, `ended` fires, `endSong("complete")` runs, and the next song starts.

**Measurements**

Four stalls across two source formats, all on 1.23.0:

| Song | Source | Reported duration | Playback froze at | Short of media end |
|---|---|---|---|---|
| A | MP3+CDG | 245 | 236.743 | 1.257s (playlist 238.00) |
| B | MP3+CDG | 266 | 248.544 | 1.496s (playlist 250.04) |
| C | YouTube MP4 | 257 | ~256 | ~1s |
| D | MP3+CDG | 251 | 241.905 | playlist not captured |

In every case the HLS playlist was complete and well formed, `#EXT-X-ENDLIST` present, no gaps. `now_playing_position` froze to the byte across repeated samples 5 seconds apart, with `is_paused: false` and no ffmpeg process running, so the transcode had finished and the player had simply stopped.

Nothing was logged server side, which is correct behaviour: the server was waiting for a report the browser was never going to send.

**The track mismatch, measured directly**

[...]

Issue #928, opened 2026-09-02
[BUG] End a song the server can see has finished


safepay:
**Why:** a host whose splash screen stalls gets a party that stops between songs and stays stopped until someone finds the machine and presses skip. Afterwards the queue advances on the server's own clock, whatever the browser is doing.

## Problem

- `is_playing` clears in exactly one place, `PlaybackController.end_song()`, and that is reached only from the master splash's `socket.emit("end_song")`. The run loop starts the next song only when `is_playing` is `False`, so one missed report stops the queue for the night. The server has no timeout of its own.
- The score screen can hold it open forever. `endSong()` awaits `startScore()`, which ends at score.js:33 returning a promise resolved only by `applause.onended` - no timeout, no `catch`, no `error` listener. If the applause never plays or never ends, the overlay stays up, the emit never happens, and the queue is wedged.
- A backgrounded tab makes this routine rather than rare: `requestAnimationFrame` stops entirely while the tab is hidden and timers clamp to roughly one a minute, so the reporter is exactly the participant the operating system is allowed to suspend.
- A stalled `<video>` emits nothing - not `ended`, not `error`, not reliably `stalled` - so nothing on the client can notice it. #942 measures four such stalls on Chromium 150+, each frozen about a second short of a stream that is complete behind it. What the server can see is `now_playing_position` unchanged across repeated samples.

## Solution

- Race the applause promise against a timeout and add an `error` listener, so the score screen can never hold the queue open.
- End a playback server-side when no `playback_position` has arrived for long enough, rather than when its reported duration has elapsed. `now_playing_duration` is probed from the source file, so for CDG it overshoots what the stream contains by the whole audio/video gap - #942 measured 266 reported against 250.04 delivered, a position that can never reach its duration. #945 narrows that gap without closing it, and a wall-clock key also catches a master that has stopped reporting entirely.
- This composes with the playback-id work rather than duplicating it: the splash already stops and clears when the server reports nothing playing, so a server-driven end recovers the screen as well as the queue.

## Done when

- [ ] A song ends without any client saying so, once its position has stopped moving
- [ ] The score screen cannot hold the queue open when its audio does not play
- [ ] A splash left in a background tab still advances the queue
- [ ] A splash frozen mid-song, emitting no events, still advances the queue

GeekShopAdmin:
[...]

PR #945, opened 2026-09-08
fix: hold the last CDG frame so video is not shorter than audio

GeekShopAdmin:
**Why:** CDG graphics end 7–17s before the music, so the transcode leaves video shorter than audio and songs lose their endings — a 245s song plays 238s. Part of #942.

## Changes

- `tpad` clones the last CDG graphic frame, and `-t` trims the output to the audio length. Measured gap: 7.10s → 0.258s.
- `-t` uses the source duration plus `avsync`, unrounded. `adelay`/`atrim` change the audio length after the probe, so the raw source duration would cut ~2s off every song when `--avsync` is set, and `get_media_duration`'s rounding would clip ~0.09s off the 245.088s example.
- `get_media_duration` now returns a float; `FileResolver` keeps the rounded value for the UI and an exact one for the trim.
- Both are gated on `is_cdg` and travel together — `tpad` leaves the video unbounded, so without the trim the output would run 600s past the music. `-t` is omitted entirely if the probe fails.

**This does not on its own stop the queue wedging.** The residual gap is frame quantisation at 25fps, still enough for Chromium 150+ to stop short, so a stall timeout is needed as well (#928).

## Test plan

- [x] Full-length CDG playback on a Pi 4, final lyric screen held through the outro — with the earlier `-shortest` form of this change.
- [x] Built command verified for CDG, non-CDG, `avsync` ±2, and a failed probe: `-t` present only for CDG, and never without a duration.
- [x] End-to-end ffmpeg run on 6.1.1: `-t 22.035918` → video 22.040 / audio 22.023.
- [ ] `-t` form on hardware — not yet run.
- [ ] `--streaming-format mp4`, which now shares the same options — not exercised.

### DIFF
--- a/pikaraoke/lib/ffmpeg.py
+++ b/pikaraoke/lib/ffmpeg.py
@@ -15,18 +15,18 @@ if TYPE_CHECKING:
     from pikaraoke.lib.file_resolver import FileResolver
 
 
-def get_media_duration(file_path: str) -> int | None:
+def get_media_duration(file_path: str) -> float | None:
     """Get the duration of a media file in seconds.
 
     Args:
         file_path: Path to the media file.
 
     Returns:
-        Duration in seconds (rounded), or None if unable to determine.
+        Duration in seconds, or None if unable to determine.
     """
     try:
         duration = ffmpeg.probe(file_path)["format"]["duration"]
-        return round(float(duration))
+        return float(duration)
     except:
         return None
 
@@ -113,17 +113,18 @@ def build_ffmpeg_cmd(
         logging.info("Playing CDG/MP3 file: " + fr.file_path)
         cdg_input = ffmpeg.input(fr.cdg_file_path, copyts=None)
         video = cdg_input.video.filter("fps", fps=25)
-        # CDG graphics routinely end seconds before the music does: the lyrics finish
-        # and the instrumental outro plays on with nothing left to draw. That leaves
-        # video shorter than audio -- 7 to 17s on the discs I have measured -- and a
-        # browser that stops at the shorter track then never fires "ended", so the
-        # queue never advances. Hold the last graphic frame instead; -shortest below
-        # trims the output at the end of the audio, so the whole song still plays.
+        # CDG graphics end 7-17s before the music, and Chromium 150+ stops at the
+        # shorter track rather than playing through, so "ended" never fires (#942).
+        # 600s is a ceiling on any outro; -t trims back to the audio length.
         video = video.filter("tpad", stop_mode="clone", stop_duration=600)
         if cdg_pixel_scaling:
             video = video.filter("scale", -1, 720, flags="neighbor")
     else:
         video = input.video
+
+    cdg_opts = {"pix_fmt": "yuv420p"}
+    if is_cdg and fr.duration_exact:
+        cdg_opts["t"] = fr.duration_exact + avsync
 
     # Build output based on format
     if force_mp4_encoding:
@@ -141,7 +142,7 @@ def build_ffmpeg_cmd(
             f="mp4",
             video_bitrate=vbitrate,
             movflags=movflags,
-            **({"pix_fmt": "yuv420p", "shortest": None} if is_cdg else {}),
+            **(cdg_opts if is_cdg else {}),
         )
     else:
         # HLS format with fMP4 segments
@@ -166,8 +167,7 @@ def build_ffmpeg_cmd(
             hls_fmp4_init_filename=fr.init_filename,
             hls_segment_filename=fr.segment_pattern,
             video_bitrate=vbitrate,
-            # CDG needs pix_fmt for proper color space
-            **({"pix_fmt": "yuv420p", "shortest": None} if is_cdg else {}),
+            **(cdg_opts if is_cdg else {}),
             **{
                 "fps_mode": "cfr",
                 "avoid_negative_ts": "make_zero",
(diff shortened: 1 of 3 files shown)


safepay:
Nice find, and thanks for the measurements - getting the outros back makes this worth having on its own.

One thing before it goes in: I can't reproduce the tight `-shortest` cut on a newer ffmpeg. On a 2026-02 git build, 20s of audio against 12s of video through your exact filter chain:

```
-shortest   video 26.12 / audio 20.00
-shortest   video 26.20 / audio 20.00
-shortest   video 25.88 / audio 20.00
-t 20       video 20.000 / audio 20.000   (3/3)
```

Same story on the HLS branch (playlist summed 25.64s), and `-shortest_buf_duration` doesn't rein it in. `-shortest` fires at the muxer, after frames are already through the filter graph, and `tpad` leaves nothing bounding how far ahead the video runs - hence 0.258s on your 5.1.9 and ~6s here, varying run to run.

So the mismatch ends up reversed rather than closed: audio becomes the shorter track. That's still a win, and the bigger one - the whole song is heard now - but it leaves a held frame sitting there for a few seconds.

Could you use `-t` instead? `FileResolver` already probes the source duration, so the value's to hand. One wrinkle: `get_media_duration` rounds to whole seconds, which would clip ~0.09s off your 245.088s example - and since you measured 0.258s as still enough to stall, an unrounded probe might be worth folding into the same change.

Last small thing: could the comment come down a bit? Most of it is reasoning that won't be news to anyone reading the filter, and the `-shortest` sentence would be stale if you take the `-t` suggestion. What I'd keep is the measured gap and the browser quirk, since neither is recoverable from the code - plus a word on where 600 comes from, which is the one number a reader will stop on:

```python
# CDG graphics end 7-17s before the music, and Chromium 150+ stops at the shorter
# track rather than playing through, so "ended" never fires (#942). 600s is a
# ceiling on any outro; the output is trimmed back to the audio length below.
```

GeekShopAdmin:
Taking the comment as you've written it, and switching to `-t`. Two things from re-testing, one of which changes the value you'd want to pass.

**I can't reproduce the overrun on 6.1.1.** Your exact shape — 20s audio, 12s video, `fps=25,tpad=stop_mode=clone:stop_duration=600`, `-fps_mode cfr`:

```
-shortest, mp4, 5 runs        video 20.000 / audio 20.000    identical 5/5
-shortest, hls fmp4, 3 runs   playlist 20.000                identical 3/3
```

Adding `copyts` on the video input, which is what `build_ffmpeg_cmd` actually does: video 20.000 / audio 20.023, again identical run to run.

So that's three builds and three behaviours — 5.1.8 leaves 0.258s, 6.1.1 lands exact, your 2026-02 build overruns ~6s with variance between runs. Which I think is a better argument for `-t` than the one I started with: `-shortest` isn't wrong in principle, it just isn't dependable across builds. Agreed, switching.

**But `fr.duration` is the wrong value to pass.** You're right about the rounding — `get_media_duration` returns `round(float(...))`, so `-t 245` clips 0.088s off the 245.088s example. The bigger problem is that the audio chain changes the duration before it reaches the muxer, while `fr.duration` is probed from the untouched source:

| audio chain (measured, 6.1.1) | 20.036s source becomes |
|---|---|
| `loudnorm` — **on by default** | 20.000 |
| `adelay` (`avsync > 0`) | 22.000 |
| `atrim` (`avsync < 0`) | 18.023 |

With `--avsync 2` and `-t 20.036`, the output audio measured 20.049s for a song that is 22.0s long. Two seconds gone off the end of every song, silently.

`-t (duration + avsync)` measures video 22.040 / audio 22.023, so that's the value — with an unrounded probe. I'd add a float helper rather than change `get_media_duration`'s return type, since `fr.duration` also feeds `now_playing_duration` in the UI. `loudnorm`'s ~36ms comes off in the safe direction, so it doesn't need handling.

**One question, because it decides what this PR actually fixes.** You describe the reversed mismatch as still a win — the whole song is heard — and I'm not sure the queue agrees. MSE's buffered range is the intersection of the two tracks while `duration` reports the longer one, so with audio as the shorter track I'd expect playback to stop at the audio end and never reach `duration`, meaning `ended` still doesn't fire and the queue still holds — just after the last note rather than before it.

That is reasoning, not measurement, and I want to be clear about which. I couldn't test it: the only Chromium I can script here is 141, which predates the regression window entirely. The one hard data point I have points that way — on 5.1.8 with `tpad` applied and only 0.258s of residual mismatch, the box still logged `Reason: stalled` and the queue still held.

So, on your 2026-02 build with the ~6s overrun: did `ended` fire and the next song start, or did it play to the end of the audio and then sit on the held frame? If it sat, then this PR restores the outro but doesn't close #942 on its own, and the stall watchdog is carrying the queue half regardless of what the transcode does. Worth establishing before it merges, so it isn't merged expecting the hang to go away.

safepay:
PR Comments and code comments are way too long. Please review comments and trim to specifics before posting. Claude is bloating everything.

safepay:
Specifically, always ask Claude to simplify the code and always use the /code-review function.

Read the comments yourself - if they are too verbose and overcomplicated for you, they are for eveyone else too!

GeekShopAdmin:
Fair — 18 lines of comment for 15 of code across the three. Pushed: your comment verbatim, `-t` instead of `-shortest`, and the duplicated splat collapsed into one dict (which also closes the mp4 path getting `tpad` without a trim). PR body cut to Why/Changes/Test plan.

Two corrections worth having:

`-t fr.duration` cuts the song short whenever `--avsync` is set — `adelay`/`atrim` change the audio length after the probe. Measured `--avsync 2` with `-t 20.036`: 20.049s of audio on a 22.0s song. Now `duration + avsync`, unrounded, and `-t` omitted entirely if the probe fails.

`test_ffmpeg` and `test_youtube_dl` both needed updating — #943 was failing two tests before this, which its checks never showed because the workflow is awaiting approval. All three branches now pass 1404 tests and black.

